PR #58, opened 2021-01-23
Generated files now have a '.g.cs' extension

odalet:
A 2-character long PR: I simply modified the filename pattern so that generated files end in `.g.cs` and not simply in `.cs`. This is useful when integrating with 3rd-party code analysis tools. This way, one can easily exclude generated files from such analyses.


AArnott (approved):
Thanks for the idea and PR.

odalet:
You're welcome!

Issue #65, opened 2021-01-24
Comments in NativeMethods.txt


CBenghi:
It would be helpful to add comments to the NativeMethods.txt file to make the list of lines more expressive.

For example, something like:
```
// GetLastInputInfo is used to determine if the user is idle in module xxx
GetLastInputInfo
// todo: remove SetWindowsHookEx and use class yyy instead
SetWindowsHookEx
```

Having read the code 

https://github.com/microsoft/CsWin32/blob/d2f2d7fb2ac3f6a68cde49f3798a32de79edb073/src/Microsoft.Windows.CsWin32/SourceGenerator.cs#L116-L123

this seems easy... I'll post a PR shortly.

PR #66, opened 2021-01-24
Added comment parsing check for  NativeMethods.txt

CBenghi:
Lines starting with "//" will be ignored when parsing NativeMethods.txt
to allow more expressive configuration files.

I've also documented the option for comments in the readme file.

This is a potential (simplistic) solution to #65.


AArnott (approved):
Thanks!

Issue #23, opened 2021-01-14
Touch-up whitespace issues


AArnott:
- [x] Some methods get a blank line between them while others do not.
- [x] https://github.com/dotnet/roslyn/issues/49733 NormalizeWhitespace should put `*` next to the type rather than the identifier
- [x] SafeHandle-derived types' `Null` property looks terrible

PR #140, opened 2021-02-23
New codegen for inline array fields

### LINKED ISSUE #117: Avoid using `Span<T>` on TFMs older than net45
This causes the project to fail to build for any target frameworks.

For example, I'm targeting a library to `net35;net48;net5.0-windows`.

jnm2:
Closes #117

TODO:

- [ ] Add range checks to the fallback indexer implementation to parallel span range checks
- [x] Ask JaredPar to review the generated code and see if `ref p[index]` should be `ref Unsafe.AsRef<uint>(p + index)`

Currently generated code (odd whitespace is not my fault; tracked by https://github.com/microsoft/CsWin32/issues/23):

```cs
[StructLayout(LayoutKind.Sequential, Pack = 1)]
internal struct __dwReserved_4
{
    internal uint _1, _2, _3, _4;
    internal ref uint this[int index]
    {
        get
        {
            unsafe
            {
                fixed (uint *p = &_1)
                    return ref p[index];
            }
        }
    }
}
```

@AArnott I like keeping `unsafe` scoped as tightly as possible, meaning I use the statement form if the signature itself isn't unsafe, but let me know if this is too weird or something, or if using an embedded statement instead of a block inside the `fixed` statement is a problem.


jnm2:
@AArnott I had to take care of some discoveries that I only made by looking at CI results. I think this is the last update I'll need to get CI passing, but regardless, the "ready for review" status is accurate. There are outstanding questions in the linked issue.

jnm2:
Nope, one more.

> error SA1633: The file header is missing or not located at the top of the file.

jnm2:
Oh, quick question. Do you know if source generators output files containing CRLF on Linux by design, or is this unusual?

```
  Error Message:
   Assert.Equal() Failure
          ↓ (pos 0)
Expected: \n    internal partial struct MainAVIHeade···
Actual:   \r\n    internal partial struct MainAVIHead···
          ↑ (pos 0)
```

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -377,6 +371,78 @@ internal enum FILE_CREATE_FLAGS
...
+        internal __dwReserved_4 dwReserved;
+        [StructLayout(LayoutKind.Sequential, Pack = 1)]
+        internal struct __dwReserved_4
+        {
+            internal uint _1, _2, _3, _4;
+            internal ref uint this[int index] => ref AsSpan()[index];
+            internal Span<uint> AsSpan() => MemoryMarshal.CreateSpan(ref _1, 4);
+        }
+    }
+";
```
AArnott, line 440: I'm not a fan of character-level equality checks as I expect they will cost too much to maintain. Rather, specific API traits can be asserted as other tests in this class do, and for functional testing, see `BasicTests.cs`. But we can keep this if you like, at least until it requires too many updates. :) It certainly helps in a PR review to see what you expect.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -377,6 +371,78 @@ internal enum FILE_CREATE_FLAGS
...
+        internal uint dwPaddingGranularity;
+        internal uint dwFlags;
+        internal uint dwTotalFrames;
+        internal uint dwInitialFrames;
+        internal uint dwStreams;
+        internal uint dwSuggestedBufferSize;
+        internal uint dwWidth;
+        internal uint dwHeight;
+        internal __dwReserved_4 dwReserved;
+        [StructLayout(LayoutKind.Sequential, Pack = 1)]
```
AArnott, line 392: @tannergooding were you saying we don't need to specify the pack size?

test/Microsoft.Windows.CsWin32.Tests/TestUtils.cs
```diff
@@ -0,0 +1,18 @@
...
+    using System.Text.RegularExpressions;
+
+    internal static class TestUtils
+    {
+        private const string ExpectedGeneratedSourceLineEnding = "\r\n";
+        private static readonly Regex NewLineRegex = new(@"\r?\n");
+
+        public static string NormalizeToExpectedLineEndings(string text)
+        {
+            return NewLineRegex.Replace(text, ExpectedGeneratedSourceLineEnding);
```
AArnott, line 15: line endings on multi-line string checks can be tough in git. If this works though, it's good enough for me.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
tannergooding: Right. Inline arrays are just like regular arrays and the natural padding of a struct is correct.

tannergooding: (assuming the metadata doesn't specify otherwise)

AArnott:
I've asked JaredPar on another thread to review and will report back here.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
jnm2: Saw that too. I can add this change to this PR, no problem. :+1:

test/Microsoft.Windows.CsWin32.Tests/TestUtils.cs
jnm2: Yep, it does work.

AArnott:
Just an update: I am actively working on this and have lots of changes coming. I'm just working to get all tests passing. The new extension methods we have to declare bring on some new challenges.

jnm2:
@AArnott Thanks for completing this!

AArnott:
Thanks for starting it, @jnm2! Particularly because it led us to discover bugs in how we did it before.

PR #157, opened 2021-02-26
Add extension methods so more runtimes can index into inline arrays

### LINKED ISSUE #152: Add extension methods to help index into inline arrays on downlevel TFMs
This carries forward what #140 tried to do but grew out of control given our current low-fidelity idea of how to qualify type names in generated code.

AArnott:
Closes #152

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -227,6 +227,7 @@ namespace Microsoft.Windows.CsWin32
         private readonly PEReader peReader;
         private readonly MetadataReader mr;
         private readonly SignatureTypeProvider signatureTypeProvider;
+        private readonly SignatureTypeProvider signatureTypeProviderFullyQualified;
         private readonly SignatureTypeProvider signatureTypeProviderAlwaysUseIntPtr;
         private readonly SignatureTypeProvider signatureTypeProviderNoMarshaledTypes;
         private readonly SignatureTypeProvider signatureTypeProviderNoMarshaledTypesOrNint;
@@ -307,10 +308,11 @@ namespace Microsoft.Windows.CsWin32
             this.peReader = new PEReader(this.metadataStream);
             this.mr = this.peReader.GetMetadataReader();
 
-            this.signatureTypeProvider = new SignatureTypeProvider(this, preferNativeInt: this.LanguageVersion >= LanguageVersion.CSharp9, preferMarshaledTypes: true);
-            this.signatureTypeProviderAlwaysUseIntPtr = new SignatureTypeProvider(this, preferNativeInt: false, preferMarshaledTypes: true);
-            this.signatureTypeProviderNoMarshaledTypes = new SignatureTypeProvider(this, preferNativeInt: true, preferMarshaledTypes: false);
-            this.signatureTypeProviderNoMarshaledTypesOrNint = new SignatureTypeProvider(this, preferNativeInt: false, preferMarshaledTypes: false);
+            this.signatureTypeProvider = new SignatureTypeProvider(this, preferNativeInt: this.LanguageVersion >= LanguageVersion.CSharp9, preferMarshaledTypes: true, fullyQualifiedNames: false);
+            this.signatureTypeProviderFullyQualified = new SignatureTypeProvider(this, preferNativeInt: this.LanguageVersion >= LanguageVersion.CSharp9, preferMarshaledTypes: true, fullyQualifiedNames: true);
+            this.signatureTypeProviderAlwaysUseIntPtr = new SignatureTypeProvider(this, preferNativeInt: false, preferMarshaledTypes: true, fullyQualifiedNames: false);
+            this.signatureTypeProviderNoMarshaledTypes = new SignatureTypeProvider(this, preferNativeInt: true, preferMarshaledTypes: false, fullyQualifiedNames: false);
+            this.signatureTypeProviderNoMarshaledTypesOrNint = new SignatureTypeProvider(this, preferNativeInt: false, preferMarshaledTypes: false, fullyQualifiedNames: false);
             this.customAttributeTypeProvider = new CustomAttributeTypeProvider();
 
             this.Apis = this.mr.TypeDefinitions.Select(this.mr.GetTypeDefinition).Where(td => this.mr.StringComparer.Equals(td.Name, "Apis")).ToList();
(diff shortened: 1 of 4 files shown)


AArnott:
CC: @jnm2

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -3154,48 +3176,43 @@ private ParameterSyntax CreateParameter(MethodSignature<TypeSyntax> methodSignat
...
+                ParameterSyntax thisParameter = Parameter(atThis.Identifier).WithType(QualifiedName(this.GetQualifiedName(fieldDef.GetDeclaringType()), fixedLengthStructName)).AddModifiers(Token(SyntaxKind.ThisKeyword));
+                ParameterSyntax indexParameter = Parameter(indexParamName.Identifier).WithType(PredefinedType(Token(SyntaxKind.IntKeyword)));
+                SyntaxTokenList methodModifiers = TokenList(Token(this.Visibility), Token(SyntaxKind.StaticKeyword), Token(SyntaxKind.UnsafeKeyword));
+                MethodDeclarationSyntax getAtMethod = MethodDeclaration(RefType(qualifiedElementType).WithReadOnlyKeyword(Token(SyntaxKind.ReadOnlyKeyword)), "GetAt")
+                    .WithModifiers(methodModifiers)
+                    .AddParameterListParameters(thisParameter.AddModifiers(Token(SyntaxKind.InKeyword)), indexParameter)
+                    .WithBody(body);
+                this.inlineArrayIndexerExtensionsMembers.Add(getAtMethod);
+
+                ////internal static unsafe ref uint GetOrSetAt(this ref MainAVIHeader.__dwReserved_4 @this, int index)
```
jnm2, line 3206: Could this be named `ItemRef` to match the BCL? https://github.com/dotnet/corefx/pull/25738

AArnott: Good idea. I was hurting for a good name.

jnm2: I just saw `GetAt` and realized that they can't both be called `ItemRef`.

AArnott: Right. I have to use unique names. I wonder which should be called `ItemRef`. In your PR, those are all immutable collections, but the name `ItemRef` doesn't *suggest* it is immutable.

AArnott: I'm inclined to go with `ItemRef` on the read/write ref. But what to rename `GetAt` to? `ItemReadRef`?

jnm2: `ItemRef`/`ReadOnlyItemRef` is a mouthful. Which of the two will be typed most often?

AArnott: The r/w will be used most often. If someone actually stores one of these structs in a `readonly` field, that's when they'll be forced to use the readonly method.

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -3137,7 +3159,7 @@ private ParameterSyntax CreateParameter(MethodSignature<TypeSyntax> methodSignat
                     //     internal Span<TheStruct> AsSpan() => MemoryMarshal.CreateSpan(ref _1, 4);
```
jnm2, line 3159: Can `internal ref TheStruct this[int index] => ref AsSpan()[index];` be made an extension method too, so that the current C# compiler tracks the ref lifetime correctly?

In fact, why not always emit the same extension method impl for it whether or not AsSpan exists?

src/Microsoft.Windows.CsWin32/Generator.cs
jnm2: I refreshed the page and your last three comments appeared. Sorry about that! I like the direction you went with it.

src/Microsoft.Windows.CsWin32/Generator.cs
AArnott: No, proper indexers cannot be written as extension methods AFAIK. 
This indexer only exists where a `Span<T>` member can legitimately exist as well, so the ref lifetime should be fine, shouldn't it?

> In fact, why not always emit the same extension method impl for it whether or not AsSpan exists?

We do. Notice the tests which show the extension methods are always generated. On netcoreapp3.1+ the extension methods are not necessary, but I keep them so folks who multitarget can write using just the extension methods so it's just one codebase.

jnm2: My understanding was that the Span member is not legitimate and can't be, from the standpoint of tracking the lifetime of the ref. We fixed the GC hole issue, but we intentionally left the issue that the span can outlive the memory it points to. Same problem with the indexer.

jnm2: From https://github.com/microsoft/CsWin32/issues/117#issuecomment-785192927

jnm2: So not having an indexer and having the C#-9-safe extension method seems better to me because it doesn't come with the outliving problem that the indexer and the AsSpan member bring.

jnm2: If the indexer and `AsSpan` method have clear XML documentation that explains that the ref can live longer than the memory it points to, and be left pointing at arbitrary memory, then I don't mind them. `CreateUnsafeSpan` might be a better name too.

AArnott: This seems out of scope of the PR since since it only adds extension methods. But I will ask Jared about it and open a new PR if fixing the existing instance members is required. Thanks for calling it out.

jnm2: > We do. Notice the tests which show the extension methods are always generated. On netcoreapp3.1+ the extension methods are not necessary, but I keep them so folks who multitarget can write using just the extension methods so it's just one codebase.

This answers my main concern. I missed that.

If the indexer and `AsSpan` method have clear XML documentation that explains that the ref and the span can live longer than the memory they point to, and be left pointing at arbitrary memory if you don't keep the target alive, then I don't mind them. `CreateUnsafeSpan` might be a better name too.

What if you make the indexer get and set directly instead of returning a ref? That way you have the nice syntax for the 99% case, but you don't have the lifetime mismatch.

jnm2: In other words:

```cs
internal TheStruct this[int index]
{
    get => this.ItemRef(index);
    set => this.ItemRef(index) = value;
}
```

PR #158, opened 2021-02-26
Remove AsSpan from inline array struct

AArnott:
This was vulnerable to accessing memory after the stack it belonged to was destroyed:

```cs
var x = new __dwReserved_4();
return ref x[0];
```

If C# 10 gets https://github.com/dotnet/csharplang/blob/master/proposals/low-level-struct-improvements.md, then we can annotate it as ThisRefEscapes and it becomes safe.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -286,7 +286,6 @@ namespace Microsoft.Windows.CsWin32
         private readonly GeneratorOptions options;
         private readonly CSharpCompilation? compilation;
         private readonly CSharpParseOptions? parseOptions;
-        private readonly bool canCallCreateSpan;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Generator"/> class.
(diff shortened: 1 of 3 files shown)


AArnott:
CC: @jnm2

jnm2:
👍 What did you think about providing an indexer that returns a struct copy to the caller rather than a ref that the compiler can't track? That way the syntax `x.inlineArray[1] = 2`; still works.

```
internal TheStruct this[int index]
{
    get => this.ItemRef(index);
    set => this.ItemRef(index) = value;
}
```

tannergooding (approved):
LGTM.

It might be worth considering an "unsafe" switch for users to explicitly opt into this given the benefits, but with the nice big warning label about why its unsafe.

AArnott:
@tannergooding It's just the "stack hole" that we're facing here, right? If we can explain that well enough, maybe we keep the AsSpan method but call it `UnsafeAsSpan` and explain the risk in the api doc. 

@jnm2 My concern about the non-ref indexer is that this *wouldn't* work:

```cs
struct.inline[3].Foo = 3;
```

AArnott:
After another chat with Jared and @tannergooding, I'm going to keep these members and add api documentation to them warning them of the "stack hole" case.

PR #159, opened 2021-02-26
Add doc comments to "unsafe" `AsSpan()` and indexer members

AArnott:


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -82,6 +82,20 @@ public class Generator : IDisposable
 
         private const string SimpleFileNameAnnotation = "SimpleFileName";
         private const string OriginalDelegateAnnotation = "OriginalDelegate";
+        private static readonly SyntaxTriviaList InlineArrayUnsafeAsSpanComment = ParseLeadingTrivia(@"/// <summary>
+/// Gets this inline array as a span.
+/// </summary>
+/// <remarks>
+/// Important: When this struct is on the stack, do not let the returned span outlive the stack frame that defines it.
```
jnm2, line 89: I sometimes use the '⚠' character in XML docs.

AArnott: Sounds good.

AArnott:
oh nuts. The multi-line text comparison is rearing its ugly head. Test fails on linux due to inconsistent line endings.

jnm2:
...a good thing the tests are there, right? Otherwise you might be generating source with inconsistent line endings. 😁

PR #172, opened 2021-03-03
Add Length property to inline arrays

### LINKED ISSUE #169: Length property on inline arrays
PCWSTR has a Length property. Do you see the potential for there to be value in adding one to a fixed-length inline array type? What about variable-length inline arrays, when the metadata supports them?

jnm2:
Closes #169

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -96,6 +96,9 @@ namespace Microsoft.Windows.CsWin32
 /// </summary>
 ");
 
+        private static readonly XmlTextSyntax DocCommentStart = XmlText(" ").WithLeadingTrivia(DocumentationCommentExterior("///"));
+        private static readonly XmlTextSyntax DocCommentEnd = XmlText(XmlTextNewLine("\r\n", continueXmlDocumentationComment: false));
+
         private static readonly IdentifierNameSyntax ConstantsClassName = IdentifierName("Constants");
         private static readonly IdentifierNameSyntax InlineArrayIndexerExtensionsClassName = IdentifierName("InlineArrayIndexerExtensions");
         private static readonly TypeSyntax SafeHandleTypeSyntax = IdentifierName("SafeHandle");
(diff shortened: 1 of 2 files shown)


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -446,6 +448,8 @@ internal partial struct MainAVIHeader
         internal struct __dwReserved_4
         {
             internal uint _0, _1, _2, _3;
+            /// <summary>Always <c>4</c>.</summary>
+            internal int Length => 4;
```
AArnott, line 452: Lets make it a constant rather than a property. That may unblock compiler optimizations.

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -3158,17 +3161,34 @@ private ParameterSyntax CreateParameter(MethodSignature<TypeSyntax> methodSignat
...
+                            .WithSemicolonToken(Token(SyntaxKind.SemicolonToken))
+                            .WithLeadingTrivia(Trivia(DocumentationCommentTrivia(SyntaxKind.SingleLineDocumentationCommentTrivia).AddContent(
+                                DocCommentStart,
+                                XmlElement("summary", List(new XmlNodeSyntax[]
+                                {
+                                    XmlText("Always "),
+                                    XmlElement("c", List(new XmlNodeSyntax[] { XmlText(lengthLiteralSyntax.Token.ValueText) })),
+                                    XmlText("."),
+                                })),
+                                DocCommentEnd))));
```
AArnott, line 3191: You might not need the doc comment once we make it a constant, since I think Intellisense will show the constant's value.
Also, the precedent for doc comments is to write it out mostly as a string literal and ask the OM to parse it, since as we see here, this is super-complicated code for a comment.

AArnott (changes requested):
Thanks!

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
jnm2: How would you use a constant, syntactically?

AArnott: It wouldn't change the syntax for usage at all. I'm merely suggesting we change this:

```diff
-internal int Length => 4;
+internal const int Length = 4;
```

So callers are unaffected, but it can now be used in more places (e.g. attributes) and compilers or the JIT may optimize based on knowing the value never changes.

jnm2: I understand that part about constants, but don't constants and static members preclude the syntax `header.dwReserved.Length` by requiring qualification by the class name instead of instance invocation?

jnm2: On phone, but I checked sharplab just to be sure I wasn't totally blanking somehow and it gave me "error CS0176: Member cannot be accessed with an instance reference; qualify it with a type name instead."

AArnott: Oh, I see what you're getting at. I'd forgotten about that impact. Ok, we can keep what you've got.

AArnott (approved):

PR #185, opened 2021-03-10
Update metadata to 10.0.19041.5-preview.68

AArnott:


jnm2:
https://github.com/microsoft/win32metadata/issues/251 was just merged which adds AlsoUsableForAttribute. Is that worth waiting for?

AArnott:
@jnm2: It's high time we update, so I don't want to delay any further. We can update this dependency as often as it's pushed to nuget.org though (the PR you reference isn't even available on nuget.org yet), so we can pick it up soon.

PR #210, opened 2021-03-17
Omit `DefaultDllImportSearchPathsAttribute` generation on older TFMs

### LINKED ISSUE #208: Please skip emitting DefaultDllImportSearchPaths on targets < net45
This seems like an easy fix which would be in line with past PRs that keep CsWin32 usable for projects that include a net35 target.

https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.dllimportsearchpath#applies-to

Alternatively, an additional source file could be generated internally defining DefaultDllImportSearchPathsAttribute and DllImportSearchPath. This would have value *if* (1) an assembly compiled for .NET Framework 3.5 was loaded into .NET Framework 4.5+ or .NET Core and (2) the runtime recognized the attributes regardless of which assembly defined them. I don't know about (2) without testing. (1) is fairly likely. For instance, I have a library targeting `net35;net48;net5.0-windows`. If someone uses it in a `net472` project, they will get the `net35` build.

AArnott:
When the attribute isn't defined, don't emit it.
This also adds a basic sanity test to ensure that at least the most basic extern method generated works on net40.

Fixes #208

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -323,6 +323,7 @@ namespace Microsoft.Windows.CsWin32
         private readonly bool canCallCreateSpan;
         private readonly bool generateSupportedOSPlatformAttributes;
         private readonly bool generateSupportedOSPlatformAttributesOnInterfaces; // only supported on net6.0 (https://github.com/dotnet/runtime/pull/48838)
+        private readonly bool generateDefaultDllImportSearchPathsAttribute;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Generator"/> class.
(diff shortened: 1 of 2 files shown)


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -339,6 +340,7 @@ public Generator(Stream metadataLibraryStream, GeneratorOptions? options = null,
             this.parseOptions = parseOptions;
 
             this.canCallCreateSpan = this.compilation?.GetTypeByMetadataName(typeof(MemoryMarshal).FullName)?.GetMembers("CreateSpan").Any() is true;
+            this.generateDefaultDllImportSearchPathsAttribute = this.compilation?.GetTypeByMetadataName(typeof(DefaultDllImportSearchPathsAttribute).FullName) is object;
```
jnm2, line 343: Drive-by comment: `is not null` is the new `is object`

AArnott: Yes, I've heard of it. My habits haven't changed yet though. :)

PR #223, opened 2021-03-24
 Update metadata to 10.0.19041.5-preview.101

AraHaan:
Updated the metadata package.

There may be more changes needed however.

See https://github.com/microsoft/win32metadata/releases/tag/v10.0.19041.5-preview.101 for the release notes to the update.


AArnott (approved):
Looks good. Given all tests pass, we'll go with it.

PR #241, opened 2021-04-05
Add `[SupportedOSPlatform]` to friendly overloads

### LINKED ISSUE #229: Friendly overloads are missing SupportedOSPlatform
## Actual behavior

```cs
    internal static unsafe int GetStagedPackagePathByFullName2(string packageFullName, PackagePathType packagePathType, ref uint pathLength, PWSTR path)
    {
        fixed (uint *pathLengthLocal = &pathLength)
        {
            fixed (char *packageFullNameLocal = packageFullName)
            {
                int __result = PInvoke.GetStagedPackagePathByFullName2(packageFullNameLocal, packagePathType, pathLengthLocal, path);
                return __result;
            }
        }
    }

    [DllImport("api-ms-win-appmodel-runtime-l1-1-3", ExactSpelling = true)]
    [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
    [SupportedOSPlatform("windows10.0.10240")]
    internal static extern unsafe int GetStagedPackagePathByFullName2(PCWSTR packageFullName, PackagePathType packagePathType, uint *pathLength, PWSTR path);
```

## Expected behavior

The `SupportedOSPlatform` attribute should appear on the friendly overload as well or the analyzer can't do its work.

## Repro steps

```
GetStagedPackagePathByFullName2
```

### Context

- CsWin32 version: 0.1.435-beta
- Target Framework: `net5.0-windows7.0;net472;netstandard2.0`
- `LangVersion` (if explicitly set by project): `9`

AArnott:
Fixes #229

### DIFF
--- a/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
+++ b/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
@@ -108,6 +108,18 @@ public class GeneratorTests : IDisposable, IAsyncLifetime
         }
     }
 
+    [Fact]
+    public void SupportedOSPlatform_AppearsOnFriendlyOverloads()
+    {
+        const string methodName = "GetStagedPackagePathByFullName2";
+        this.compilation = this.starterCompilations["net5.0"];
+        this.generator = new Generator(this.metadataStream, DefaultTestGeneratorOptions, this.compilation, this.parseOptions);
+        Assert.True(this.generator.TryGenerate(methodName, CancellationToken.None));
+        this.CollectGeneratedCode(this.generator);
+        this.AssertNoDiagnostics();
+        Assert.All(this.FindGeneratedMethod(methodName), method => Assert.Contains(method.AttributeLists, al => IsAttributePresent(al, "SupportedOSPlatform")));
+    }
+
     [Theory, PairwiseData]
     public void COMInterfaceWithSupportedOSPlatform(bool net50, bool allowMarshaling)
     {
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -3683,7 +3683,6 @@ namespace Microsoft.Windows.CsWin32
                     .WithModifiers(modifiers)
                     .WithAttributeLists(List<AttributeListSyntax>())
                     .WithParameterList(ParameterList().AddParameters(parameters.ToArray()))
-                    .WithLeadingTrivia(leadingTrivia)
                     .WithBody(body)
                     .WithSemicolonToken(default);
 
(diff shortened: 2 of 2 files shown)


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -108,6 +108,18 @@ public void SimplestMethod(string tfm)
...
+    public void SupportedOSPlatform_AppearsOnFriendlyOverloads()
+    {
+        const string methodName = "GetStagedPackagePathByFullName2";
+        this.compilation = this.starterCompilations["net5.0"];
+        this.generator = new Generator(this.metadataStream, DefaultTestGeneratorOptions, this.compilation, this.parseOptions);
+        Assert.True(this.generator.TryGenerate(methodName, CancellationToken.None));
+        this.CollectGeneratedCode(this.generator);
+        this.AssertNoDiagnostics();
+        Assert.All(this.FindGeneratedMethod(methodName), method => Assert.Contains(method.AttributeLists, al => IsAttributePresent(al, "SupportedOSPlatform")));
+    }
```
jnm2, line 121: What happens on `netstandard2.0` or `net4.8`? Ideal would be for SupportedOSPlatformAttribute to be defined as internal.

AArnott: The point of this attribute is to allow an analyzer to compare against the TFM, and AFAIK only the net5.0 TFM supports mixing in the Windows version. So even if we did define the attribute ourselves, would the analyzer even be able to do its work?

AArnott: I'm merging the PR because the change you propose is outside the scope of this PR anyway. But if you reply I'll still read it.

jnm2: Ah, I guess you'd have to declare `<TargetPlatformIdentifier>` yourself. Good point, not likely someone would do that and I haven't checked how the .NET 5 SDK analyzers behave in this situation when targeting netcoreapp3.1.

My main concern is whether the generated code is checking for the availability of this attribute in the compilation before emitting attribute syntax though. I didn't spot the check.

jnm2: Ah, found the check:

https://github.com/microsoft/CsWin32/blob/fe62ed485959c67e2ae5f899b8cc316010276af9/src/Microsoft.Windows.CsWin32/Generator.cs#L345

AArnott: And then there's the surprising bit that the attribute isn't allowed on interfaces till .NET 6, which we also handle:

https://github.com/microsoft/CsWin32/blob/fe62ed485959c67e2ae5f899b8cc316010276af9/src/Microsoft.Windows.CsWin32/Generator.cs#L350-L350

PR #249, opened 2021-04-28
Added missing tests to kernel32 and DbgHelp.

### LINKED ISSUE #224: Missing tests for MINIDUMP_TYPE, MiniDumpWriteDump, GetCurrentProcess, GetCurrentProcessId, GetCurrentThreadId.
There may be other apis, however there are missing tests that would check generation on all 3 of the TFMs the tests try to generate for.

If this gets tested it will expose possibly where the issue might be for #221 on these apis. Oddly enough it happens with the enum as well for some unknown reason that I think should generate the enum on .NET Standard 2.0 (but that is another issue for the one I linked) and yes I tried every version of CsWin32 and they all failed the same way on these.

AraHaan:
Fixes #224

(Tests created on my Mac)

### DIFF
--- a/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
+++ b/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
@@ -200,7 +200,10 @@ public class GeneratorTests : IDisposable, IAsyncLifetime
             "ISpellCheckerFactory", // COM interface that includes `ref` parameters
             "LocalSystemTimeToLocalFileTime", // small step
             "ID3D12Resource", // COM interface with base types
-            "ID2D1RectangleGeometry")] // COM interface with base types
+            "ID2D1RectangleGeometry", // COM interface with base types
+            "GetCurrentProcess", // A handle for the current process
+            "GetCurrentProcessId", // An ID of the current process
+            "GetCurrentThreadId")] // An ID of the current thread of the current process
         string api,
         bool allowMarshaling)
     {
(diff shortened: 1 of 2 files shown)


AraHaan:
Wait why doesn't the pipeline test ns2.0 as well so it can expose the failures on these when targeting that TFM?

AArnott:
> why doesn't the pipeline test ns2.0 as well

netstandard2.0 isn't a runtime. It's just a standard. Tests can only run on a runtime like .NET Core, .NET Framework, or mono.
It looks like we only run tests on .NET 5, regardless of OS. For the code generation tests that seems adequate since I do not expect the runtime to influence the code generation itself. But we could add more runtimes if necessary.

AArnott:
We don't want a test for every API to make sure it generated. We have tests for specific APIs that tend to be problematic and throw exceptions during generation, but ultimately those are redundant with our "generate everything" tests. Why do we need these new tests?

AraHaan:
Because a lot of those api's I added are technically some of the common ones I find that people want to use on the internet, myself included.

AraHaan:
I see this is why I have my issue on my codebase not generating this enum:
```
  Failed GeneratorTests.GetMINIDUMP_TYPEGeneration [126 ms]
  Error Message:
   Assert.True() Failure
Expected: True
Actual:   False
  Stack Trace:
     at GeneratorTests.GetMINIDUMP_TYPEGeneration() in D:\a\1\s\test\Microsoft.Windows.CsWin32.Tests\GeneratorTests.cs:line 248
```

Which is odd because it should be constant across all cpu specific builds of Windows 10 that has the latest dbghelp.dll.

AArnott:
> Because a lot of those api's I added are technically some of the common ones I find that people want to use on the internet, myself included.

To be clear, I'm not arguing that we shouldn't generate those APIs. We do. I'm just saying they already have test coverage as part of the full-generation test. If we have a bug where certain APIs are simply *missing* in the generated code, then that's certainly worth taking a look at. It sounds like that's what you're saying, so I'll take another look.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -229,6 +232,33 @@ public void GetLastErrorNotIncludedInBulkGeneration()
...
+    /// Verifies that the MINIDUMP_TYPE enum is generated.
+    /// </summary>
+    // TODO: Figure out why this enum is not being generated and get it to generate as it should be constant with every platform.
+    [Fact]
+    public void GetMINIDUMP_TYPEGeneration()
+    {
+        this.generator = new Generator(this.metadataStream, DefaultTestGeneratorOptions, this.compilation, this.parseOptions);
+        Assert.True(this.generator.TryGenerate("MINIDUMP_TYPE", CancellationToken.None));
+        this.CollectGeneratedCode(this.generator);
+        Assert.True(this.generator.TryGetEnumName("MINIDUMP_TYPE", out string? actualDeclaringEnum));
```
AArnott, line 245: You're using the wrong API here. `TryGetEnumName` can't be used to verify that the enum is generated. 
By adding `this.AssertNoDiagnostics();` to the test the generated code is logged as test output and I see that the enum is in fact generated. So no bug here, it seems.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -229,6 +232,33 @@ public void GetLastErrorNotIncludedInBulkGeneration()
...
+        Assert.Equal("MINIDUMP_TYPE", actualDeclaringEnum);
+    }
+
+    /// <summary>
+    ///  Verifies that MiniDumpWriteDump is generated.
+    /// </summary>
+    // TODO: Make test try to generate using x86, x64, arm, and arm64 instead of AnyCPU so that it passes.
```
AArnott, line 252: The trick here is to add this line to the top of your test:

```cs
        this.compilation = this.compilation.WithOptions(this.compilation.Options.WithPlatform(Platform.X86));
```

With that, the test passes. So again, apparently no bug here.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -200,7 +200,10 @@ public void COMInterfaceWithSupportedOSPlatform(bool net50, bool allowMarshaling
             "ISpellCheckerFactory", // COM interface that includes `ref` parameters
             "LocalSystemTimeToLocalFileTime", // small step
             "ID3D12Resource", // COM interface with base types
-            "ID2D1RectangleGeometry")] // COM interface with base types
+            "ID2D1RectangleGeometry", // COM interface with base types
+            "GetCurrentProcess", // A handle for the current process
+            "GetCurrentProcessId", // An ID of the current process
+            "GetCurrentThreadId")] // An ID of the current thread of the current process
```
AArnott, line 206: I don't see the value of testing these individually unless they provide a faster path to exposing an existing test failure in the `FullGeneration` test.

Issue #199, opened 2021-03-16
Leverage `AlsoUsableForAttribute` for more interoperable typedefs


AArnott:
https://github.com/microsoft/win32metadata/issues/251 was just merged which adds AlsoUsableForAttribute. Is that worth waiting for?

_Originally posted by @jnm2 in https://github.com/microsoft/CsWin32/issues/185#issuecomment-800652291_

Also to accept SafeHandle everywhere that accepts the base type struct (e.g. HGDIOBJ, as mentioned in #205).

AArnott:
An interesting sanity check after this feature is implemented would be to see if this sample compiles without errors:

```cs
using System.Drawing;
using Windows.Win32.Foundation;
using Windows.Win32.Graphics.Gdi;
using static Windows.Win32.PInvoke;

class Helper
{
    public Bitmap GetScreenCapture(HWND handle)
    {
        Bitmap bitmap;

        RECT region;
        GetWindowRect(handle, out region);

        HWND desktopWindowHandle = GetDesktopWindow();
        HDC windowDCHandle = GetWindowDC(desktopWindowHandle);
        CreatedHDC memoryDCHandle = CreateCompatibleDC(windowDCHandle);
        HBITMAP bitmapHandle = CreateCompatibleBitmap(windowDCHandle, region.right - region.left, region.bottom - region.top);
        HGDIOBJ bitmapOldHandle = SelectObject(memoryDCHandle, bitmapHandle);

        bool success = BitBlt(memoryDCHandle, 0, 0, region.right - region.left, region.bottom - region.top, windowDCHandle, region.left, region.top, ROP_CODE.SRCCOPY| ROP_CODE.CAPTUREBLT);

        try
        {
            bitmap = Image.FromHbitmap(bitmapHandle);
        }
        finally
        {
            SelectObject(memoryDCHandle, bitmapOldHandle);
            DeleteObject(bitmapHandle);
            DeleteDC(memoryDCHandle);
            ReleaseDC(desktopWindowHandle, windowDCHandle);
        }

        return bitmap;
    }
}
```

Thanks to https://github.com/dotnet/pinvoke/issues/593 who provided this sample.

AArnott:
See also #491 which may make sense to fix at the same time as this one.

PR #258, opened 2021-05-02
Add IsNull property on pointer-sized structs

### LINKED ISSUE #206: Please add `.IsNull` helper on handle structs to improve the usage syntax over `.Value == 0`
Instead of one of these:
```cs
if (PInvoke.SelectObject(bitmapDC, objectHandle).Value == 0)
```
```cs
if (PInvoke.SelectObject(bitmapDC, objectHandle).Equals(0))
```

I'd like to do something with the word 'null' in it since it's a handle:
```cs
if (PInvoke.SelectObject(bitmapDC, objectHandle).IsNull)
```

Alternatives that I would not prefer are ` == HGDIOBJ.Null` or `== default`.

jnm2:
Closes #206. Specifics may depend on the conversation there.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -2828,7 +2828,7 @@ namespace Microsoft.Windows.CsWin32
                 return this.DeclareTypeDefBOOLStruct(typeDef);
             }
 
-            bool isHandle = false;
+            bool isHandle = name.Identifier.ValueText == "HGDIOBJ";
             foreach (CustomAttributeHandle attHandle in typeDef.GetCustomAttributes())
             {
                 CustomAttribute att = this.mr.GetCustomAttribute(attHandle);
@@ -2874,7 +2874,8 @@ namespace Microsoft.Windows.CsWin32
                 .WithExpressionBody(ArrowExpressionClause(AssignmentExpression(SyntaxKind.SimpleAssignmentExpression, fieldAccessExpression, valueParameter)))
                 .WithSemicolonToken(Token(SyntaxKind.SemicolonToken)));
 
-            if (fieldInfo.FieldType is IdentifierNameSyntax { Identifier: { Value: "nint" or "nuint" or nameof(IntPtr) or nameof(UIntPtr) } })
+            // If this typedef struct represents a pointer, add an IsNull property.
+            if (fieldInfo.FieldType is IdentifierNameSyntax { Identifier: { Value: nameof(IntPtr) or nameof(UIntPtr) } })
             {
                 // internal static bool IsNull => value == default;
                 members = members.Add(PropertyDeclaration(PredefinedType(Token(SyntaxKind.BoolKeyword)), "IsNull")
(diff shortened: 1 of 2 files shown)


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -2874,6 +2874,15 @@ private StructDeclarationSyntax DeclareTypeDefStruct(TypeDefinition typeDef)
                 .WithExpressionBody(ArrowExpressionClause(AssignmentExpression(SyntaxKind.SimpleAssignmentExpression, fieldAccessExpression, valueParameter)))
                 .WithSemicolonToken(Token(SyntaxKind.SemicolonToken)));
 
+            if (fieldInfo.FieldType is IdentifierNameSyntax { Identifier: { Value: "nint" or "nuint" or nameof(IntPtr) or nameof(UIntPtr) } })
```
jnm2, line 2877: I expected this to be `is PrimitiveTypeSyntax { IsNint: true } or { IsNuint: true }` but I saw that PrimitiveTypeHandleInfo is constructing IdentifierName instead of PredefinedType. I couldn't see a way to create such a PrimitiveTypeSyntax though.

jnm2: Never mind, just saw the definition of IsNint. 😆
http://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Syntax/InternalSyntax/TypeSyntax.cs

AArnott, line 2877: Null is not a value for an integer. Zero is. Can we drop `nint` and `nuint` checks from this?
Or if you need it for those, what if we call it `IsZero`? (at least for the non-pointer field types)

AArnott (changes requested):

src/Microsoft.Windows.CsWin32/Generator.cs
AArnott: And I'm doubtful about the value of IsZero, since I suspect `== 0` is legal syntax for these structs.

jnm2: @AArnott HGDIOBJ has a `nint` field. I don't care at all about if `nint` works per say, but I absolutely want handles to have `IsNull`.

AArnott: I'd say the issue then is that `HGDIOBJ` uses `nint` when it should be using `IntPtr`. We may be able to work this out without hard-coding when we address #199, but until then I've hard-coded it so that your test can pass without adding `IsNull` to all `nint` structs.

AArnott (approved):

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -2828,7 +2828,7 @@ private StructDeclarationSyntax DeclareTypeDefStruct(TypeDefinition typeDef)
                 return this.DeclareTypeDefBOOLStruct(typeDef);
             }
 
-            bool isHandle = false;
+            bool isHandle = name.Identifier.ValueText == "HGDIOBJ";
```
jnm2, line 2831: Why is this name special-cased? Temporary measure until metadata takes charge of this determination?

jnm2: Ah, expanded the resolved conversation and found your answer. 👍

PR #262, opened 2021-05-07
Add dependencies to package also replaced IsExternalInit with nuget package providing the same thing.

### LINKED ISSUE #261: Include transitive closure of dependencies in package
...sticking a debugger onto the Omnisharp process, it appears the generator is throwing an exception. We also seem to be eating the exception rather than reporting it, so we'll need to fix that! But the exception is:

    Could not load file or assembly 'System.Numerics.Vectors, Version=4.1.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified.

Which seems to point to the first problem: the generator package is packaging dependencies like System.Text.Json, but isn't packaging dependencies of System.Text.Json (namely System.Numerics.Vectors):

https://github.com/microsoft/CsWin32/blob/e67e9c55652afa1139409790622c1176e730db02/src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.targets#L23-L35

So that probably should be updated to copy in the dependency. 

_Originally posted by @jasonmalinowski in https://github.com/microsoft/CsWin32/discussions/218#discussioncomment-696590_

AraHaan:
Also I updated the sdk-api submodule because the docs changed since then I think the submodule was probably forgotten in terms of updates so I updated it.

Fixes #261.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,0 +1,838 @@
+NOTICES AND INFORMATION
+Do Not Translate or Localize
+
+This software incorporates material from third parties.
+Microsoft makes certain open source code available at https://3rdpartysource.microsoft.com,
+or you may send a check or money order for US $5.00, including the product name,
+the open source component name, platform, and version number, to:
+
+Source Code Compliance Team
+Microsoft Corporation
+One Microsoft Way
+Redmond, WA 98052
+USA
+
+Notwithstanding any other terms, you may reverse engineer this software to the extent
+required to debug changes to any libraries licensed under the GNU Lesser General Public License.
+
+---------------------------------------------------------
+
+NuGet.Frameworks 5.6.0 - Apache-2.0
+
+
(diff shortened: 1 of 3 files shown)


src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.csproj
```diff
@@ -48,6 +48,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.CodeAnalysis.CSharp" Version="3.8.0" />
     <PackageReference Include="Nullable" Version="1.3.0" />
+    <PackageReference Include="IsExternalInit" Version="1.0.0" />
```
AArnott, line 51: Please revert the `IsExternalInit` change. That package is buggy.
https://github.com/manuelroemer/IsExternalInit/issues/5

PR #305, opened 2021-06-18
Fix memory leak in testing

AArnott:
Fixes a regression introduced by 8c6685a6dacd

### DIFF
--- a/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
+++ b/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
@@ -2,7 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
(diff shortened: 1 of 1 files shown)


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -22,6 +22,7 @@
 
 public class GeneratorTests : IDisposable, IAsyncLifetime
 {
+    private static readonly ReferenceAssemblies NetStandard20 = ReferenceAssemblies.NetStandard.NetStandard20.AddPackages(ImmutableArray.Create(new PackageIdentity("System.Memory", "4.5.4")));
     private static readonly ConcurrentDictionary<ReferenceAssemblies, Task<ImmutableArray<MetadataReference>>> TestReferencesCache = new();
```
AArnott, line 26: We should remove this, as per @sharwell pointing out this cache already exists internally. I just wasn't always benefitting from it till `NetStandard20` got cached and reused as I'm now doing in this PR.

PR #309, opened 2021-06-25
Consume API docs from win32metadata

AArnott:
This removes our own production of the API docs data file and replaces it with a consumption of what the win32metadata repo started producing in https://github.com/microsoft/win32metadata/pull/548.


mikebattista:
@kennykerr @marler8997 @timsneath this PR updated C#/Win32 to use the new [Microsoft.Windows.SDK.Win32Docs](https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Docs/0.1.4-alpha) package to provide Intellisense sourced from our documentation. The package uses [MessagePack](https://msgpack.org/) as the format which supports parsers for many languages including yours (just search for `msgpack-<LANG>` on GitHub).

https://github.com/microsoft/win32metadata/pull/548 and https://github.com/microsoft/win32metadata/pull/548/files#diff-db362395805d75d149824342c0c4117c5475893dccaff9fae75bf18337a8fd84 go into a bit more detail about the package and the schema. This PR serves as a reference implementation.

If you get a chance, would appreciate any feedback you could give on how your experience is consuming it to add better documentation/Intellisense to your projections. Is MessagePack an appropriate format? Is it easy to parse using the existing parsers that exist for your language? Do you have any performance concerns?

Let us know. Thanks!

marler8997:
I haven't used MessagePack, but it looks like a reasonable format for something like this.  Thanks for letting me know about this new package, I'll look into including it and report any issues I run into.

timsneath:
Excited about the potential here, for sure. From a Dart perspective, this today comes from hand-scraping from Windows docs, which is obviously laborious. I'll take a look. Does this include constants like `WM_PAINT` and structs, as well as functions?

kennykerr:
Agreed, this sounds promising. As I [mentioned here though](https://github.com/microsoft/windows-rs/issues/931#issuecomment-870845229), my main concern is general adoption of this format for APIs beyond Win32.

mikebattista:
The Win32 docs are scraped from the .md files in https://github.com/MicrosoftDocs/sdk-api/tree/docs/sdk-api-src/content and converted into MessagePack to basically build a hash table to make it easy to resolve the docs for a given API from any language.

WinRT docs are available in a similar location at https://github.com/MicrosoftDocs/winrt-api. If we can validate the approach with the Win32 docs and all agree that MessagePack is a good format here that's easy to consume from all projections, we can produce packages for other areas as well using the same model. The request for feedback though is to validate the approach first before we invest too much in this direction.

> Does this include constants like WM_PAINT and structs, as well as functions?

@aarnott can you answer this?

AArnott:
@timsneath I just checked and it looks like `WM_PAINT` does *not* have docs associated with it. But as [the docs exist](https://docs.microsoft.com/en-us/windows/win32/gdi/wm-paint) in what looks like a similar structure to functions. I filed https://github.com/microsoft/win32metadata/issues/563 to track covering this and other constants.

PR #317, opened 2021-07-06
Add tests and fix a few whitespace issues

sharwell:
* Add full tests for `BOOL` and `ReleaseDC`
* Fix indentation of constructors and operators
* Add newline after attribute list


sharwell:
@AArnott this change doesn't fix all whitespace issues, but it dramatically improves readability relative to the 0.1.506-beta release. 😄

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -808,6 +811,1175 @@ public void ProjectReferenceBetweenTwoGeneratingProjects(bool internalsVisibleTo
...
+		}
+	}
+}
+".Replace("\r\n", "\n")),
+                },
+            },
+        }.RunAsync();
+    }
+
     private static ImmutableArray<Diagnostic> FilterDiagnostics(ImmutableArray<Diagnostic> diagnostics) => diagnostics.Where(d => d.Severity > DiagnosticSeverity.Hidden).ToImmutableArray();
```
AArnott, line 1984: Find this line and remove it, since you're now demonstrating a way.

```cs
// CONSIDER: How can I pass in the source generator itself, with AdditionalFiles, so I'm exercising that code too?
```

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -808,6 +811,1175 @@ public void ProjectReferenceBetweenTwoGeneratingProjects(bool internalsVisibleTo
         this.AssertNoDiagnostics();
     }
 
+    [Fact]
```
AArnott, line 814: How should I think about the tests your adding here, which use verbatim text assertions compared to my other tests that mostly assert specific things? I can imagine a *few* verbatim tests, but each new one creates a maintenance burden when either the generator or the metadata behind it changes. I see you added *several*.

AArnott (approved):
Love it. I also see that your PR makes things faster:

image

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
sharwell: One option is to extract the output **BOOL.g.cs** to a shared constant, so if it changes you can change all cases at once.

Another option is to generate one or more outputs by running the generator with `<EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>` ([example](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/a559b3e58af09387aadc1372a504f9208bce4909/StyleCop.Analyzers/StyleCop.Analyzers/StyleCop.Analyzers.csproj#L26-L29)), and then reading files from that output location during tests. Changes would automatically apply to all tests using that output, and the diffs would be visible in source control.

PR #319, opened 2021-07-06
Generate outputs

sharwell:
This change uses the CsWin32 source generator at build time to generate the outputs that will later be expected by unit tests. In its current form, it's only usable for tests with a fixed input configuration, but that still reduces the number of cases where generated outputs are added directly to string literals (which then need to be manually maintained).

In the future, satellite projects could be used to generate test inputs under fixed configurations with similar results, or the test implementation could be updated to write expected outputs back to disk when the tests are executed.


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -1284,127 +1045,9 @@ public async Task TestSimpleMethod()
...
-		/// </remarks>
-		[DllImport(""User32"", ExactSpelling = true)]
-		[DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
-		internal static extern int ReleaseDC(win32.Foundation.HWND hWnd, win32.Graphics.Gdi.HDC hDC);
-	}
-}
-".Replace("\r\n", "\n")),
+                    LoadExpectedOutput("HDC.g.cs"),
+                    LoadExpectedOutput("HWND.g.cs"),
+                    LoadExpectedOutput("PInvoke.User32.g.cs"),
```
sharwell, line 1050: 📝 These files are the highest risk of churn. This one will work as long as _either_ no other method from user32.dll is added to NativeMethods.txt _or_ the test itself is updated to include more methods than just ReleaseDC in the test.

AArnott:
Could you add a PR description? I'm struggling to see why we'd want to check these in, and how you came to decide on _these_ files/APIs as opposed to others.

test/Microsoft.Windows.CsWin32.Tests/Microsoft.Windows.CsWin32.Tests.csproj
```diff
@@ -20,10 +32,17 @@
...
 
   <ItemGroup>
-    <ProjectReference Include="..\..\src\Microsoft.Windows.CsWin32\Microsoft.Windows.CsWin32.csproj" />
+    <ProjectReference Include="..\..\src\Microsoft.Windows.CsWin32\Microsoft.Windows.CsWin32.csproj" OutputItemType="Analyzer" Private="true" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\System.Text.Json.dll" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\Microsoft.Bcl.AsyncInterfaces.dll" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\System.Text.Encodings.Web.dll" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\Microsoft.Windows.SDK.Win32Docs.dll" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\MessagePack.dll" />
+    <Analyzer Include="$(RepoRootPath)\bin\Microsoft.Windows.CsWin32\$(Configuration)\netstandard2.0\MessagePack.Annotations.dll" />
```
AArnott, line 41: I'm concerned about making this test project also act as a source generator recipient. We already have a separate test project for that, and while it's nice to more easily see how it works, it also fails to update generated code in the IDE to reflect recent source generator changes because the IDE only loads the source generator once. 
*This* test project doesn't suffer from that because it only runs the source analyzer directly as part of tests. This feels like its muddying the waters somewhat.

test/Microsoft.Windows.CsWin32.Tests/NativeMethods.txt
```diff
@@ -0,0 +1,7 @@
+// Methods (avoid creating more than one per native DLL, as it can cause test churn)
```
AArnott, line 1: I'm wondering how with just these few items we end up with many dozens of files, and how adding one more API from the same module is somehow significantly worse than adding an API from another module. 
What is the value of adding these in the first place, and by what rule should we decide on adding more or taking away what we have?

test/Microsoft.Windows.CsWin32.Tests/Microsoft.Windows.CsWin32.Tests.csproj
```diff
@@ -1,15 +1,27 @@
...
+  <Import Project="..\..\src\Microsoft.Windows.CsWin32\build\Microsoft.Windows.CsWin32.props" />
 
   <PropertyGroup>
     <TargetFrameworks>net5.0;netcoreapp3.1;net472</TargetFrameworks>
     <RootNamespace />
   </PropertyGroup>
 
+  <PropertyGroup>
+    <EmitCompilerGeneratedFiles>true</EmitCompilerGeneratedFiles>
+    <CompilerGeneratedFilesOutputPath>$(MSBuildProjectDirectory)\.generated\$(TargetFramework)</CompilerGeneratedFilesOutputPath>
```
AArnott, line 11: FWIW: the source generator may customize its output by more dimensions than just TFM. And the TFMs this test project happens to build with doesn't necessarily represent all possibilities either. 
Tying the test runtime itself to the source generation feels a bit like an inappropriate coupling, as a single build of the test assembly should be able to test all possible outputs of the SG as run by that runtime. For example, a source generator running on `net472` should be able to generate source in a project targeting `net6.0`. This test project today can do that, but this generated code thing that you're adding wouldn't.

AArnott: I wonder if we should instead have tests have a switchable read/write mode, where they emit the generated files to their source directory when run under "update LKG" mode, and otherwise they read from those same locations.
That might help avoid the risk you bring up of cross-test contamination from NativeMethods.txt changes if each test still gets to manage its own LKG files.

AArnott: > In the future, satellite projects could be used to generate test inputs under fixed configurations with similar results, or the test implementation could be updated to write expected outputs back to disk when the tests are executed.

Sounds like you had similar ideas.

PR #331, opened 2021-07-13
Fix SA1629 "... should end with a period" in generated xml docs

georg-jung:
Probably not adding too much value, but easy to change too :)...

Also, \<remarks>\<para>... is generated like:

    /// <para><see href="https://docs.microsoft.com/windows/win32/api/iphlpapi/nf-iphlpapi-sendarp">Learn more about this API from docs.microsoft.com.</see></para>

while the end of \<summary> is generated as:

    /// <para><see href="https://docs.microsoft.com/windows/win32/api/iphlpapi/nf-iphlpapi-sendarp#parameters">Read more on docs.microsoft.com</see>.</para>

thus, beeing SA1629 compliant.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -2020,7 +2020,7 @@ namespace Microsoft.Windows.CsWin32
                     else if (docs.HelpLink is object)
                     {
                         docCommentsBuilder.AppendLine();
-                        docCommentsBuilder.AppendLine($@"/// <para><see href=""{docs.HelpLink}"">Learn more about this API from docs.microsoft.com.</see></para>");
+                        docCommentsBuilder.AppendLine($@"/// <para><see href=""{docs.HelpLink}"">Learn more about this API from docs.microsoft.com</see>.</para>");
                         docCommentsBuilder.Append("/// ");
                     }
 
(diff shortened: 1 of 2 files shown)


jnm2:
This reverses #295 and is visually irritating to me. I've been thinking about contributing a fix for SA1629. Either way, SA1629 doesn't run on generated source from what I remember.

georg-jung:
I think you're right, I didn't notice it directly on generated source code, but when experimenting with an editable copy of the generated code directly in a project. Where the `.` should be is probably a question of taste - I don't have strong feelings about it. I just sent the PR because I randomly noticed that the code that is generated doesn't stick to the default StyleCop ruleset and it's easy to fix.

Either way, we currently have the point inside the <see> in one place and outside in the other (see above).

jnm2:
Nice, I would have liked to have fix that if I had seen it.

AArnott:
Thanks for the submission. I typically strive for StyleCop compliance in the generated code. Even though those analyzers don't typically run on generated code, copying from generated code into the user's project is a supported scenario, and we don't want that to require a ton of fixups after in order to fix stylecop warnings.
That said, I know we're probably far from that sort-of goal (of being stylecop compliant). 

I'd generally prioritize avoiding analyzer warnings over catering to what @jnm2 finds "visually irritating" though (no offense intended). So I think we should take this. If @jnm2 succeeds in getting a change made to SA1929 such that either way will not produce a warning, I guess we can switch back at that point.

jnm2:
Hey, no worries. The fact that no one else seems to be bothered by the period being a different color is a sign that I should consider not switching things back even after SA1929 is updated.

jnm2:
Hey, `<a href` instead of `<see href` would not have had this conflict with SA1629. It's also shorter.

PR #357, opened 2021-08-12
Fix marshaling of NativeArrayInfo parameters

### LINKED ISSUE #269: Calling ID2D1Factory.CreateGeometryGroup causes Access violation
## Actual behavior

When I call `ID2D1Factory.CreateGeometryGroup` with a `geometriesCount` other than `0`, I get the following access violation:

```
Exception thrown at 0x00007FF92F4C8F08 (d2d1.dll) in MyApp.exe: 0xC0000005: Access violation reading location 0x0000000802400011.
```

I suspect the problem has to do with the `ID2D1Geometry[] geometries` argument not being marshaled correctly. The geometries I pass to `CreateGeometryGroup` are all valid and can be used fine on their own.

## Expected behavior

I expect it not to cause an access violation.

## Repro steps

```c#
var Factory = PInvoke.D2D1CreateFactory ( ... );
var Geoms = new ID2D1Geometry[] { ... };
Factory.CreateGeometryGroup ( D2D1_FILL_MODE.D2D1_FILL_MODE_ALTERNATE, Geoms, (uint) Geoms.Length, out var D2DGeometry );
```

```
D2D1CreateFactory
ID2D1Factory
ID2D1RenderTarget
D2D1_RENDER_TARGET_PROPERTIES
D2D1_HWND_RENDER_TARGET_PROPERTIES
D2D_RECT_F
D2D1_ROUNDED_RECT
D2D1_ELLIPSE
D2D1_FILL_MODE
D2D_MATRIX_3X2_F
D2D1_STROKE_STYLE_PROPERTIES
ID2D1RectangleGeometry
CWPSTRUCT
```

```json
{
  "public": false,
  "allowMarshaling": true
}
```

### Context

- CsWin32 version: `0.4.422-beta`
- Win32Metadata version (if explicitly set by project): `10.0.19041.202-preview`
- Target Framework: `net5.0`
- `LangVersion`: n/a

pziezio:
Fixes #269 and many other problems with DirectX2D/11/12 etc.

When parameters with `NativeArrayInfo` are marshaled `MarshalAsAttribute` is sometimes missing.

https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs#L32
In this case `marshalAs` variable corresponds to element type of the array.
When there is no need for setting specific `ArraySubType` the `MarshalAsAttribute` is dropped altogether.
It should be set to just `MarshalAs(UnmanagedType.LPArray)` without specifying `ArraySubType`.

https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs#L57
There is exactly the same code in case of marshaling structs as array. Maybe that caused some confusion.
But in that case dropping `MarshalAs` is correct.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs
+++ b/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Windows.CsWin32
                     // But this pointer represents an array, so type as an array.
                     return new TypeSyntaxAndMarshaling(
                         ArrayType(elementSyntax).AddRankSpecifiers(ArrayRankSpecifier()),
-                        marshalAs is object ? new MarshalAsAttribute(UnmanagedType.LPArray) { ArraySubType = marshalAs.Value } : null);
+                        marshalAs is object ? new MarshalAsAttribute(UnmanagedType.LPArray) { ArraySubType = marshalAs.Value } : new MarshalAsAttribute(UnmanagedType.LPArray));
                 }
                 else if (xIn || xOut)
                 {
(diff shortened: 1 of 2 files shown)


AArnott (approved):
Thanks!

PR #363, opened 2021-08-17
Stop translating BOOL to bool

### LINKED ISSUE #362: GetMessage returns a bool when it should return a BOOL
## Actual behavior

Currently, `GetMessage` is defined to return a bool, which can be compared to true or false only. This incorrect according to [the documentation](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-getmessage):

> If the function retrieves a message other than WM_QUIT, the return value is nonzero.
> 
> If the function retrieves the WM_QUIT message, the return value is zero.
> 
> If there is an error, the return value is -1. For example, the function fails if hWnd is an invalid window handle or lpMsg is an invalid pointer. To get extended error information, call GetLastError.

This means, that from C#, it's impossible to handle errors for this function.

Because of the bool return, it actually ends up being that *ignoring* errors is encouraged, when the docs actually call this out as wrong:

> Because the return value can be nonzero, zero, or -1, avoid code like this:
> ```cpp
> while (GetMessage( lpMsg, hWnd, 0, 0))
> ```

## Expected behavior

GetMessage should return an `int`, so that it can be compared to -1 and 0.

## Repro steps

```
GetMessage
```

Nothing

No

### Context

- CsWin32 version: 0.1.506-beta
- Win32Metadata version (if explicitly set by project): Not explicitly set
- Target Framework: net5.0-windows
- `LangVersion` (if explicitly set by project): Not explicitly set

AArnott:
Fixes #362

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Windows.CsWin32
 
         internal static readonly Dictionary<string, TypeSyntax> AdditionalBclInteropStructsMarshaled = new Dictionary<string, TypeSyntax>(StringComparer.Ordinal)
         {
-            { "BOOL", PredefinedType(TokenWithSpace(SyntaxKind.BoolKeyword)) },
+            ////{ "BOOL", PredefinedType(TokenWithSpace(SyntaxKind.BoolKeyword)) },
         };
 
         internal static readonly Dictionary<string, TypeSyntax> BclInteropSafeHandles = new Dictionary<string, TypeSyntax>(StringComparer.Ordinal)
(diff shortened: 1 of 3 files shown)


sylveon:
I'd personally not add the implicit conversion operator, as it encourages the `while (GetMessage(...))` kind of code, which is explicitly pointed out as wrong in the docs for GetMessage. Or maybe this special case is something an analyzer could handle.

jnm2:
That would hurt usability of all APIs that use `BOOL` for the sake of one or two that are this way, right?

sylveon:
That's why I suggested adding an analyzer instead, that would warn for such usages.

AArnott:
I imagine analyzers could be a huge win for users. But also a bottomless pit of work. If folks are interested in sending PRs that add analyzers, I will try to find time to add the first one as a template for how more can be added.

PR #386, opened 2021-09-01
Enable creating C# projections based on multiple input winmd's and docs

AArnott:
**TODO**

* [x] Test with multiple metadata inputs
* [x] https://github.com/microsoft/win32metadata/pull/629 completed and the new package consumed as part of this PR.

❓ Which metadata should be brought in automatically by CsWin32? win32metadata is the only one right now. Should we start including others? Should we *remove* win32metadata and leave no metadata inputs by default to better support folks who only want to generate projections for their own APIs?


mikebattista:
For now, we should continue to pull in win32metadata but support adding NuGet packages to the project that pull in additional metadata. Project templates can add additional references to metadata as well as prepopulate NativeMethods.txt to improve their out-of-box experience.

PR #412, opened 2021-09-18
Generate code with consistent line endings

### LINKED ISSUE #369: VS editor constantly warns about inconsistent line endings
## Actual behavior

The warning showed in the below screenshot is displayed every time I open a file generated by CsWin32. Because these generated files are read-only, the options in the warning dialog do nothing, and will not stop the alert from appearing the next time I open one. This, while technically harmless, is very annoying, as I want to keep the dialog enabled so I can detect mismatched line endings in other, non-generated files. 

Screenshot

### Context

- Visual Studio version: VS 2022 Preview 3.1
- CsWin32 version: 0.1.506-beta
- Win32Metadata version: (default)
- Target Framework: net6.0-windows10.0.18362
- `LangVersion`: 10 (as required by generated code, which includes `global using` statements)

AArnott:
This worksaround VS's annoying dialog when opening generated code that offers to normalize line endings (on the readonly file).

Fixes #369

### DIFF
--- a/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
+++ b/test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
@@ -15,6 +15,7 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Testing;
+using Microsoft.CodeAnalysis.Text;
 using Microsoft.Windows.CsWin32;
 using Microsoft.Windows.CsWin32.Tests;
 using Xunit;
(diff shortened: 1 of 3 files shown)


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -2170,6 +2171,38 @@ static string JoinAssemblyMetadata(string name)
...
 
+    private static void AssertConsistentLineEndings(Compilation compilation)
+    {
+        foreach (SyntaxTree doc in compilation.SyntaxTrees)
+        {
+            AssertConsistentLineEndings(doc);
+        }
+    }
+
+    private static void AssertConsistentLineEndings(SyntaxTree syntaxTree)
```
sharwell, line 2182: 💡 You could use the new `EqualOrDiff` functionality from dotnet/roslyn-sdk for this assertion by asserting the `syntaxTree.GetText()` is the same as `syntaxTree.GetText().Replace("\r\n", "\n")`.

https://github.com/dotnet/roslyn-sdk/pull/886

AArnott: Ooh, that's nice. Thanks for the tip. Where is this `EqualOrDiff` method exposed, or would I need to copy its source?

sharwell: If you already reference MS.CA.Testing, you can use `DefaultVerifier` with [this extension method](https://github.com/dotnet/roslyn-sdk/blob/1c5cbe8d9add7cce50443620730124d60e714f2f/src/Microsoft.CodeAnalysis.Testing/Microsoft.CodeAnalysis.Analyzer.Testing/Extensions/IVerifierExtensions.cs#L30). Otherwise the source is reasonably easy to copy.

sharwell: Note that `EqualOrDiff` was originally implemented by you, but we modified it in dotnet/roslyn-sdk#878 and dotnet/roslyn-sdk#886 to specifically support cases where only the line endings changed.

AArnott: Ya, I thought I remembered that I had shared it. We developed it for vs-mef.

PR #447, opened 2021-11-09
Update win32metadata to 15.0.1-preview

AArnott:


AArnott:
Waiting on https://github.com/microsoft/win32metadata/discussions/735 to update the metadata one more time so we can revert 3168668f6dc6263b02d37bd86f5f7f5afd1adb35 before merging.

PR #450, opened 2021-11-16
Check whether Span<T> exists before using it

jnm2:
Found a few more spots.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -4747,18 +4747,21 @@ namespace Microsoft.Windows.CsWin32
 
             if (elementType is PredefinedTypeSyntax { Keyword: { RawKind: (int)SyntaxKind.CharKeyword } })
             {
-                // internal readonly bool Equals(string value) => Equals(value.AsSpan());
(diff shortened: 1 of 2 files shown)


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -4102,7 +4105,8 @@ private IEnumerable<MethodDeclarationSyntax> DeclareFriendlyOverloads(MethodDefi
                 }
                 else if ((externParam.Type is PointerTypeSyntax { ElementType: TypeSyntax ptrElementType }
                     && !IsVoid(ptrElementType)
-                    && !this.IsInterface(parameterTypeInfo)) ||
+                    && !this.IsInterface(parameterTypeInfo)
+                    && this.canUseSpan) ||
```
jnm2, line 4109: My confidence is rather low in this placement and the following one. Does this make sense to you or did I leave holes?

AArnott: I'm not sure either. We can take it as-is and fix it with more tests if anything comes up. It seems safe enough since `canUseSpan` is true in the officially supported scenarios, so it seems at worst your change would simply not cut out use of span in enough places.

AArnott (approved):

PR #540, opened 2022-05-07
README: Fix Newer metadata section commandline

Thog:
should have been ``--prerelease``, ``-pre`` is not a valid argument.


AArnott (approved):
Thanks

PR #579, opened 2022-06-05
Make `[In, Optional]` managed struct parameters actually optional in friendly overloads

### LINKED ISSUE #578: `[Optional, In]` parameter shouldn't use `in` modifier in extern method for managed structs
The `MiniDumpWriteDump` extern method is emitted with an `in` modifier for its last parameter. While convenient when specifying the struct value, it makes it syntactically impossible to pass in a null pointer to omit the value. 
The metadata describes this last parameter as `[Optional, In]`, so there should be a way to pass in a null pointer.

AArnott:
Fixes #578

### DIFF
--- a/src/Microsoft.Windows.CsWin32/ArrayTypeHandleInfo.cs
+++ b/src/Microsoft.Windows.CsWin32/ArrayTypeHandleInfo.cs
@@ -28,4 +28,6 @@ internal record ArrayTypeHandleInfo(TypeHandleInfo ElementType, ArrayShape Shape
             return new TypeSyntaxAndMarshaling(PointerType(element.Type));
         }
     }
+
+    internal override bool? IsValueType(TypeSyntaxSettings inputs) => false;
 }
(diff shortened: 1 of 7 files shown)


AArnott:
@AaronRobinsonMSFT .NET Framework interop can handle `Unsafe.AsRef<MINIDUMP_CALLBACK_INFORMATION>(null)` as an argument to an `in MINIDUMP_CALLBACK_INFORMATION` parameter of an `extern` method in order to pass in a null pointer, right? It compiles. I just want to verify that it runs.

AaronRobinsonMSFT:
> .NET Framework interop can handle Unsafe.AsRef<MINIDUMP_CALLBACK_INFORMATION>(null) as an argument to an in MINIDUMP_CALLBACK_INFORMATION parameter of an extern method in order to pass in a null pointer, right?

@AArnott Yes. This will translate into a `nullptr`/`NULL` on the native side.

For .NET 5+ we recommend using [`Unsafe.NullRef<T>`](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.unsafe.nullref), but I assume you don't have that on a .NET Standard 2.1 platform.

AArnott:
Thanks. The source generator already caters to various TFMs and C# LangVersions, so we can certainly emit `Unsafe.NullRef<T>` wherever that API is defined.

PR #587, opened 2022-06-16
Add option to generate out IntPtr for COM output pointer parameters

### LINKED ISSUE #328: Add option to generate `out IntPtr` for COM output pointer parameters

I am making heavy use of the .NET 5/6 ComWrappers class to make my Windows Forms codebases compatible with [NativeAOT](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT). I can create as many ComWrappers subclasses as I want, each of which knows how to to create bindings for a different set of COM interfaces. (I am using [this Roslyn source generator](https://github.com/kant2002/WinFormsComInterop/tree/main/WinFormsComInterop.SourceGenerator) to create the marshaling code.) However, only one ComWrappers instance can be used with `ComImport` interfaces, and it must know about every `ComImport` interface that is used, anywhere in the program and all of its libraries. There is no good way to do this _a priori_, and there is no way to specify an alternate ComWrappers instance for a `ComImport` type. The global instance must be used in this situation.

However, if I have a COM instance pointer as an IntPtr, I can use an instance of ComWrappers to create an AOT-compatible RCW for that instance. As long as I do not use `UnmanagedType.IUnknown` or any similar API, I can use whatever ComWrappers object I want, allowing me to pick one that understands the type of object I am giving it. However, there is currently no way I know of to have CsWin32 generate this kind of binding.

Add a flag in NativeMethods.json that will generate code that uses `out IntPtr` as the type for arguments annotated as `ComOutPtr`. I can then use my custom ComWrappers code to call methods, or use `Marshal.QueryInterface` and `Marshal.Release` on the pointer directly.

Calling `Marshal.GetIUnknownForObject()` is not an option, as that call will occur after the default ComWrappers instance attempts to generate an RCW for the pointer.

**Additional context**

I will use extension methods to create overloads for these APIs that are easier to call.

As a side note: It would be really, _really_ cool if there was some way to make `WinFormsComInterop.SourceGenerator` aware of CsWin32 and be able to generate marshaling stubs for all the interfaces automatically, but unfortunately you can’t use source generator output as input to another source generator. I must manually provide an attribute that triggers the WinFormsComInterop source generator for every interface used. However, one line of code per interface is far preferable to having to bind all of it myself. Thanks!

AArnott:
Closes #328

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -424,6 +424,8 @@ public class Generator : IDisposable
 
     internal SuperGenerator? SuperGenerator { get; set; }
 
+    internal GeneratorOptions Options => this.options;
+
     internal string InputAssemblyName { get; }
 
     internal MetadataIndex MetadataIndex { get; }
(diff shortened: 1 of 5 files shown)


AArnott:
CC: @wjk

PR #609, opened 2022-07-20
Trim Zero White Space

### LINKED ISSUE #608: Zero width space characters break NativeMethods.txt
## Actual behavior

When a line contains a zero width space character it breaks the build.

## Expected behavior

Zero width space characters should be ignored/removed when processing.

## Repro steps

```
ActivateKeyboardLayout
﻿BitBlt
```

String.Trim doesn't remove zero width space characters:
https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=net-6.0

> The .NET Framework 3.5 SP1 and earlier versions maintain an internal list of white-space characters that this method trims. Starting with the .NET Framework 4, the method trims all Unicode white-space characters (that is, characters that produce a true return value when they are passed to the [IsWhiteSpace(Char)](https://docs.microsoft.com/en-us/dotnet/api/system.char.iswhitespace?view=net-6.0#system-char-iswhitespace(system-char)) method). Because of this change, the [Trim()](https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=net-6.0#system-string-trim) method in the .NET Framework 3.5 SP1 and earlier versions removes two characters, ZERO WIDTH SPACE (U+200B) and ZERO WIDTH NO-BREAK SPACE (U+FEFF), that the [Trim()](https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=net-6.0#system-string-trim) method in the .NET Framework 4and later versions does not remove. In addition, the [Trim()](https://docs.microsoft.com/en-us/dotnet/api/system.string.trim?view=net-6.0#system-string-trim) method in the .NET Framework 3.5 SP1 and earlier versions does not trim three Unicode white-space characters: MONGOLIAN VOWEL SEPARATOR (U+180E), NARROW NO-BREAK SPACE (U+202F), and MEDIUM MATHEMATICAL SPACE (U+205F).

elachlan:
Fixes #608

### DIFF
--- a/src/Microsoft.Windows.CsWin32/SourceGenerator.cs
+++ b/src/Microsoft.Windows.CsWin32/SourceGenerator.cs
@@ -125,6 +125,7 @@ public class SourceGenerator : ISourceGenerator
 
     private const string NativeMethodsTxtAdditionalFileName = "NativeMethods.txt";
     private const string NativeMethodsJsonAdditionalFileName = "NativeMethods.json";
+    private static readonly char[] ZeroWhiteSpace = new char[] { '\uFEFF', '\u200B' };
 
     /// <inheritdoc/>
     public void Initialize(GeneratorInitializationContext context)
@@ -201,7 +202,7 @@ public class SourceGenerator : ISourceGenerator
                         continue;
                     }
 
-                    name = name.Trim();
+                    name = name.Trim().Trim(ZeroWhiteSpace);
                     var location = Location.Create(nativeMethodsTxtFile.Path, line.Span, nativeMethodsTxt.Lines.GetLinePositionSpan(line.Span));
                     try
                     {
--- a/src/Microsoft.Windows.CsWin32/SourceGenerator.cs
+++ b/src/Microsoft.Windows.CsWin32/SourceGenerator.cs
@@ -125,7 +125,11 @@ public class SourceGenerator : ISourceGenerator
 
     private const string NativeMethodsTxtAdditionalFileName = "NativeMethods.txt";
     private const string NativeMethodsJsonAdditionalFileName = "NativeMethods.json";
-    private static readonly char[] ZeroWhiteSpace = new char[] { '\uFEFF', '\u200B' };
+    private static readonly char[] ZeroWhiteSpace = new char[]
+    {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Windows.CsWin32/SourceGenerator.cs
```diff
@@ -125,6 +125,7 @@ public class SourceGenerator : ISourceGenerator
 
     private const string NativeMethodsTxtAdditionalFileName = "NativeMethods.txt";
     private const string NativeMethodsJsonAdditionalFileName = "NativeMethods.json";
+    private static readonly char[] ZeroWhiteSpace = new char[] { '\uFEFF', '\u200B' };
```
RussKie, line 128: The first one is BOM, the seconds one is zero width space.

elachlan: `ZERO WIDTH SPACE (U+200B)` and `ZERO WIDTH NO-BREAK SPACE (U+FEFF)`, should I add a comment documenting each one? Also should I only be checking `ZERO WIDTH SPACE (U+200B)`?

AArnott: Yes, comments with these meanings would be great.
That said, I would expect our TextReader to remove the BOM. Is it not doing that? 
And I wonder why we'd want to tolerate other strange unicode characters. We already have code elsewhere in the source generator to detect any non-ANSI character and emit a warning (or error maybe) so the user can correct it. This change looks like it would just allow these characters to creep in. Is that really a win?

elachlan: We ran into it dealing with a merge conflict via github. I think it would be helpful to sanitize the inputs as best as possible. A warning instead of an error would then pickup the ZWSP for fixing. I don't think it should stop it from compiling and it should probably let the user know its there as well.

elachlan: I can also see your point on the ZWSP. But the code already runs a Trim() against it. Which is pretty similar.

AArnott: Alright, fair enough. I'll take the PR. Can you just add the comments we talked about?

AArnott (approved):

PR #621, opened 2022-07-26
Add GeneratedCodeAttribute to Generated code

### LINKED ISSUE #594: Consider emitting GeneratedCodeAttribute on all generated types
See [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.codedom.compiler.generatedcodeattribute?f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(System.CodeDom.Compiler.GeneratedCodeAttribute)%3Bk(SolutionItemsProject)%3Bk(DevLang-csharp)%26rd%3Dtrue&view=net-6.0) for the attribute. The `LibraryImportAttribute` source generator emits code that uses this. It seems like a possibly good idea.

elachlan:
Closes #594

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -2099,33 +2099,32 @@ public class Generator : IDisposable
     private static AttributeSyntax StructLayout(TypeAttributes typeAttributes, TypeLayout layout = default, CharSet charSet = CharSet.Ansi)
     {
         LayoutKind layoutKind = (typeAttributes & TypeAttributes.ExplicitLayout) == TypeAttributes.ExplicitLayout ? LayoutKind.Explicit : LayoutKind.Sequential;
-        AttributeSyntax? structLayoutAttribute = Attribute(IdentifierName("StructLayout")).AddArgumentListArguments(
-            AttributeArgument(MemberAccessExpression(
-                SyntaxKind.SimpleMemberAccessExpression,
-                IdentifierName(nameof(LayoutKind)),
-                IdentifierName(Enum.GetName(typeof(LayoutKind), layoutKind)!))));
+        List<AttributeArgumentSyntax> args = new();
+        AttributeSyntax? structLayoutAttribute = Attribute(IdentifierName("StructLayout"));
+        args.Add(AttributeArgument(MemberAccessExpression(
+                 SyntaxKind.SimpleMemberAccessExpression,
+                 IdentifierName(nameof(LayoutKind)),
+                 IdentifierName(Enum.GetName(typeof(LayoutKind), layoutKind)!))));
 
         if (layout.PackingSize > 0)
         {
-            structLayoutAttribute = structLayoutAttribute.AddArgumentListArguments(
-                AttributeArgument(LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(layout.PackingSize)))
+            args.Add(AttributeArgument(LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(layout.PackingSize)))
                     .WithNameEquals(NameEquals(nameof(StructLayoutAttribute.Pack))));
         }
 
         if (layout.Size > 0)
         {
-            structLayoutAttribute = structLayoutAttribute.AddArgumentListArguments(
-                AttributeArgument(LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(layout.Size)))
+            args.Add(AttributeArgument(LiteralExpression(SyntaxKind.NumericLiteralExpression, Literal(layout.Size)))
                     .WithNameEquals(NameEquals(nameof(StructLayoutAttribute.Size))));
         }
 
         if (charSet != CharSet.Ansi)
         {
-            structLayoutAttribute = structLayoutAttribute.AddArgumentListArguments(
-                AttributeArgument(MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, IdentifierName(nameof(CharSet)), IdentifierName(Enum.GetName(typeof(CharSet), charSet)!)))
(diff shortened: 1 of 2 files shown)


test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -602,6 +602,40 @@ public void HandleTypeDefsUseIntPtrAsFieldType(string handleType)
...
+        StructDeclarationSyntax structDecl = Assert.IsType<StructDeclarationSyntax>(this.FindGeneratedType("INPUT_RECORD").Single());
+        Assert.Contains(structDecl.AttributeLists, al => al.Attributes.Any(a => a.Name.ToString().Contains("GeneratedCode")));
+
+        IEnumerable<ClassDeclarationSyntax> pInvokeClass = this.FindGeneratedType("PInvoke").OfType<ClassDeclarationSyntax>();
+        Assert.Contains(pInvokeClass, c => c.AttributeLists.Any(al => al.Attributes.Any(a => a.Name.ToString().Contains("GeneratedCode"))));
+
+        EnumDeclarationSyntax enumDecl = Assert.IsType<EnumDeclarationSyntax>(this.FindGeneratedType("ACTIVATE_KEYBOARD_LAYOUT_FLAGS").Single());
+        Assert.Contains(enumDecl.AttributeLists, al => al.Attributes.Any(a => a.Name.ToString().Contains("GeneratedCode")));
+
+        ClassDeclarationSyntax arrayExtensions = Assert.IsType<ClassDeclarationSyntax>(this.FindGeneratedType("InlineArrayIndexerExtensions").Single());
```
elachlan, line 632: InlineArrayIndexerExtensions and FriendlyOverloadExtensions aren't generated, I need to find an associated token which will generate them.

elachlan:
@AArnott If you could give this a quick review that would be great.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
elachlan: @AArnott this is the only outstanding problem with this issue AFAIK. Could you please give me some guidance on how I can test for this?

AArnott: I found the token to add. But now the test fails because the attribute is added above the xml doc comment for templated structs. I'm taking a look to see if I can fix it now.

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -451,13 +456,18 @@ private IEnumerable<MemberDeclarationSyntax> NamespaceMembers
     {
         get
         {
-            IEnumerable<MemberDeclarationSyntax> result =
-                from entry in this.committedCode.MembersByModule
-                select ClassDeclaration(Identifier(this.options.ClassName))
+            var first = this.committedCode.MembersByModule.Select(entry => ClassDeclaration(Identifier(this.options.ClassName))
```
AArnott, line 459: I'd like to see this change reworked so we are not repeating so much code as this was already complex. Also, as this turns out to be an expensive part CPU-wise to the generation, I don't want to use Skip(1) which will "realize" the first (and often only) element a second time. If we need to rewrite this as a loop instead of using linq extension methods, so be it.

AArnott (changes requested):
Thanks for contributing. I think you're almost there.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
elachlan: I had the same issue elsewhere and had to make sure the order of function calls was correct.

elachlan:
@AArnott, the remaining test failures are a bit of a strange one. They are all off by one tab space. 

If you call `AddAttributeLists` and then afterwards `WithLeadingTrivia`, it adds an extra tab space for some reason. I have no idea how to fix it.

AArnott:
Thank you for getting us this far. I hope to look more into this today or tomorrow.

elachlan:
@sharwell this code causes the xml comment to have an extra leading tab white space:
```csharp
typeDeclaration = typeDeclaration
                    .WithLeadingTrivia()
                    .AddAttributeLists(AttributeList().AddAttributes(GeneratedCodeAttribute))
                    .WithLeadingTrivia(typeDeclaration.GetLeadingTrivia());
```

The generated code attribute is:
```csharp
private static readonly AttributeSyntax GeneratedCodeAttribute = Attribute(IdentifierName("global::System.CodeDom.Compiler.GeneratedCode"))
        .WithArgumentList(FixTrivia(AttributeArgumentList().AddArguments(
            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(ThisAssembly.AssemblyName))),
            AttributeArgument(LiteralExpression(SyntaxKind.StringLiteralExpression, Literal(ThisAssembly.AssemblyInformationalVersion))))));
```

It seems like an issue in Roslyn, but its more likely PEBKAC. Is there anything obvious I have missed?

elachlan:
@AArnott Yousef pointed out in the roslyn issue that the problem is in `WhitespaceRewriter`. I'll let you know if I manage to solve it.

Youssef1313:
@elachlan Can you see if https://github.com/elachlan/CsWin32/pull/1 fixes the issue?

elachlan:
I'll pull yours down now and check it.

elachlan:
No It didn't fix it. I think I managed to find a way, let me push through my changes.

elachlan:
@Youssef1313 I have pushed through some more changes around for loop indenting.

This is what its generating at the moment inside `PAINTSTRUCT`
image
I am not 100% sure, but there might be braces missing.

elachlan:
@AArnott can you give this another review?

elachlan:
@Youssef1313 thank you for your help on this! It helped immensely.

AArnott (approved):
Thank you for your contribution. I would never have guessed this required so much work. I appreciate your dedication.

PR #627, opened 2022-07-27
`BOOL`/`BOOLEAN` templates are incorrect

### LINKED ISSUE #624: `BOOL`/`BOOLEAN` templates are incorrect
The `BOOL` and `BOOLEAN` template constructors taking a C# `bool` are invalid.

https://github.com/microsoft/CsWin32/blob/ab63c9ad139a77a43ad331f7334d38799d6125e2/src/Microsoft.Windows.CsWin32/templates/BOOL.cs#L3

https://github.com/microsoft/CsWin32/blob/ab63c9ad139a77a43ad331f7334d38799d6125e2/src/Microsoft.Windows.CsWin32/templates/BOOLEAN.cs#L3

The `BOOL` template should be, the `BOOLEAN` is similar.

```csharp
internal unsafe BOOL(bool value) => this.Value = value ? 1 : 0;
```

The notion of simply blitting the lowest bits of a `bool` instance is undefined. It also makes this usage troublesome if a user defines [`SkipLocalsInitAttribute`](https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.skiplocalsinitattribute), which with the current implementation leave the upper bits of `Value` in an undefined state thus making a .NET `false` possibly be non-zero.

elachlan:
Fixes #624

### DIFF
--- a/src/Microsoft.Windows.CsWin32/templates/BOOL.cs
+++ b/src/Microsoft.Windows.CsWin32/templates/BOOL.cs
@@ -1,11 +1,6 @@
 ﻿partial struct BOOL
 {
-	internal unsafe BOOL(bool value) => this.Value = value ? 1 : 0;
-	public static unsafe implicit operator bool(BOOL value)
-	{
-		sbyte v = checked((sbyte)value.Value);
-		return *(bool*)&v;
-	}
-
+	internal BOOL(bool value) => this.Value = value ? 1 : 0;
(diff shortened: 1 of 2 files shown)


AArnott:
Conversation is ongoing in the issue and should be resolved before merging this.

AArnott:
Superseded by #645. Thanks for contributing.

PR #628, opened 2022-07-28
Add drawing types templates

### LINKED ISSUE #626: Consider adding cast operators to System.Drawing types
It is tedious and error prone (particularly with RECT/Rectangle) to convert between the native types and System.Drawing types (`Point`, `Size`, `Rectangle`, and `*F` variants).

In Windows Forms we define the RECT, POINT, etc. structs with cast operators (often implicit). We also provide useful properties on RECT for convenience, notably `.Width` and `.Height`.

elachlan:
Fixes #626
Related: https://github.com/dotnet/winforms/issues/7468


elachlan:
CC: @lonitra

src/Microsoft.Windows.CsWin32/templates/RECT.cs
```diff
@@ -0,0 +1,26 @@
...
+		this(location.X, location.Y, unchecked(location.X + size.Width), unchecked(location.Y + size.Height)) { }
+	internal RECT(int left, int top, int right, int bottom)
+	{
+		this.left = left;
+		this.top = top;
+		this.right = right;
+		this.bottom = bottom;
+	}
+
+	public static RECT FromXYWH(int x, int y, int width, int height) =>
```
AArnott, line 15: Please make all non-operator members `internal`. Tey will be promoted to `public` if appropriate. Only operators are `public`, since they must be per the C# language.

test/Microsoft.Windows.CsWin32.Tests/GeneratorTests.cs
```diff
@@ -602,6 +602,21 @@ public void HandleTypeDefsUseIntPtrAsFieldType(string handleType)
...
+    [InlineData("SIZE")]
+    public void DrawingStructsGenerate(string handleType)
+    {
+        this.compilation = this.compilation.WithOptions(this.compilation.Options.WithPlatform(Platform.X64));
+        this.generator = this.CreateGenerator();
+        Assert.True(this.generator.TryGenerate(handleType, CancellationToken.None));
+        this.CollectGeneratedCode(this.generator);
+        this.AssertNoDiagnostics();
+        StructDeclarationSyntax paintStruct = Assert.IsType<StructDeclarationSyntax>(this.FindGeneratedType(handleType).Single());
+    }
```
AArnott, line 618: Thanks for the test, but we already have exhaustive tests to make sure nothing generated will produce a compiler warning or error. And testing that the type is generated when requested by name isn't something we do more than once or twice total, rather than for every type we customize. So please drop the test.

A more meaningful test would be in `BasicTests.cs`. There you can *use* the APIs you're adding via the template to confirm they are present and that they have correct implementations.

AArnott (changes requested):
Looks like you added the functionality all right where it fits the design. Thank you. Just a couple of changes, please.

src/Microsoft.Windows.CsWin32/templates/POINTF.cs
```diff
@@ -0,0 +1,11 @@
+﻿partial struct POINTF
```
elachlan, line 1: @AArnott is `POINTF` correct? The generated type seemed to live in a different namespace and I was wondering if the `*F` variants are .NET specific and not interop types.

AArnott: Yes, this looks correct. It appears there are many POINT? structs in the metadata, in several namespaces:
image

And your tests look good to.

AArnott: BTW, here's an idea I should have mentioned earlier: We have a mapping of metadata types to .NET types where they are equivalent, to promote easier interop with other .NET code. The POINT and POINTF structs look equivalent to the .NET types you're converting to and from, so maybe we should stop generating these structs altogether and just emit APIs that reference the .NET types instead. Thoughts?

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -3659,6 +3659,20 @@ private StructDeclarationSyntax DeclareStruct(TypeDefinitionHandle typeDefHandle
         // Add the additional members, taking care to not introduce redundant declarations.
         members.AddRange(additionalMembers.Where(c => c is not StructDeclarationSyntax cs || !members.OfType<StructDeclarationSyntax>().Any(m => m.Identifier.ValueText == cs.Identifier.ValueText)));
 
+        switch (name.Identifier.ValueText)
+        {
+            case "POINTF":
+            case "POINT":
+            case "RECT":
+            case "RECTF":
```
AArnott, line 3667: You mention `RECTF` and `SIZEF` here but you don't have templates for them.

src/Microsoft.Windows.CsWin32/templates/POINTF.cs
elachlan: Sure, where do I create the mapping?

src/Microsoft.Windows.CsWin32/Generator.cs
elachlan: I think SIZEF and RECTF are not metadata types. So I removed the templates.

src/Microsoft.Windows.CsWin32/templates/POINTF.cs
elachlan: This looks like where the mapping happens, I am playing around with it now.
https://github.com/microsoft/CsWin32/blob/ab63c9ad139a77a43ad331f7334d38799d6125e2/src/Microsoft.Windows.CsWin32/Generator.cs#L39-L59

elachlan: I am unsure if BannedAPIsWithoutMarshaling is required. If Marshaling is disabled, I imagine the APIs shouldn't be generated with the BCL mapping and just use the win32API Structs?

AArnott (approved):
It turns out that `SIZE` cannot be replaced with the BCL `Size` struct because the fields are in a different order. I moved POINT and POINTF to another dictionary in the generator so that they are never generated, even when not marshaling, since the BCL struct types should be adequate and this gives an even better interop experience. I deleted the point templates since we never generate those types any more.
Thank you for starting on this, for the tests and the rect and size templates. This is definitely an improvement.

Issue #91, opened 2021-02-14
GetStdHandle shouldn't return a CloseHandleSafeHandle


deiruch:
Standard output handles should rarely be closed. Here's Raymond Chen's [explanation](https://devblogs.microsoft.com/oldnewthing/20130307-00/?p=5033). Thus, the proper way to use `GetStdHandle` is by combining it with `DangerousAddRef`:
```cs
var handle = PInvoke.GetStdHandle(STD_HANDLE_TYPE.STD_OUTPUT_HANDLE);
bool ignore=default;
handle.DangerousAddRef(ref ignore);
```

This could be simplified if `GetStdHandle` wouldn't return a `CloseHandleSafeHandle`.

AArnott:
Great point. I filed https://github.com/microsoft/win32metadata/issues/244 against the metadata so we can do this along with other projections.

AArnott:
Now that #143 is done, we actually *can* create `SafeHandle` instances with `ownsHandle: false` set on them.

xPaw:
I'm curious, how are you expected to pass a handle from GetStdHandle into functions such as GetConsoleMode which expect SafeHandle?

Use SafeFileHandle or is there a better way of doing it?

```csharp
var @out = PInvoke.GetStdHandle(Windows.Win32.System.Console.STD_HANDLE.STD_OUTPUT_HANDLE);
var safeHandle = new Microsoft.Win32.SafeHandles.SafeFileHandle(@out, ownsHandle: false);
PInvoke.GetConsoleMode(safeHandle, out var mode);
```

xPaw:
Ah that's the trick, thank you.

PR #654, opened 2022-08-16
Honour DoNotRelease on SafeHandles

### LINKED ISSUE #644: GetModuleHandle shouldn't return FreeLibrarySafeHandle
## Actual behavior

`GetModuleHandle` returns `FreeLibrarySafeHandle`. Which will call `FreeLibrary` on cleanup. But, according to the documentation, this is not necessary to avoid errors:
> The `GetModuleHandle` function returns a handle to a mapped module without incrementing its reference count. However, if this handle is passed to the `FreeLibrary` function, the reference count of the mapped module will be decremented. Therefore, do not pass a handle returned by `GetModuleHandle` to the `FreeLibrary` function. Doing so can cause a DLL module to be unmapped prematurely.
https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlew#remarks

`GetModuleHandleEx` partly also affected by this problem, because the cleanup behavior depends on the arguments passed
https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandleexw#parameters

## Repro steps

```
GetModuleHandle
```

```json
```

### Context

- CsWin32 version: [e.g. `0.1.635-beta`]
- Target Framework: [e.g. `netstandard2.0`]

elachlan:
Fixes #644 and possibly #91?

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -27,6 +27,7 @@ public class Generator : IDisposable
     internal const string InteropDecorationNamespace = "Windows.Win32.Interop";
     internal const string NativeArrayInfoAttribute = "NativeArrayInfoAttribute";
     internal const string RAIIFreeAttribute = "RAIIFreeAttribute";
+    internal const string DoNotReleaseAttribute = "DoNotReleaseAttribute";
     internal const string GlobalNamespacePrefix = "global::";
     internal const string GlobalWinmdRootNamespaceAlias = "winmdroot";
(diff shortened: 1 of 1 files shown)


elachlan:
CC: @lonitra

elachlan:
CC: @AArnott

AArnott (approved):
Perfect. I wish there was a good way to test this, but I gave it a quick try and unfortunately there is no public API on `SafeHandle` that exposes the second argument passed to its constructor. And a syntax-level test would be very brittle.

PR #655, opened 2022-08-16
TypeDefs missing IsNull

### LINKED ISSUE #610: Generate HKL IsNull helper
As per https://github.com/dotnet/winforms/pull/7451#discussion_r926180298 `HKL` is missing the `IsNull` helper talked about in https://github.com/microsoft/CsWin32/issues/206

elachlan:
Fixes #610

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -3807,7 +3807,7 @@ public class Generator : IDisposable
             this.TryGenerateExternMethod(freeMethodName, out _);
         }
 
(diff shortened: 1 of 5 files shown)


elachlan:
@AArnott I feel this is the best way to fix it. I may have missed something. But it basically means typedefs do not use nint.

elachlan:
@AArnott I might need some guidance on these failures.

elachlan:
I think this approach might not be the best. I still think the meta data should output an attribute where a typedef has a value type of `DECLARE_HANDLE`.  That way we can differentiate between the types a lot easier. We could use `InvalidHandleValues` but `HWND` is an example of a typedef which doesn't have that attribute.

AArnott:
Given how hard it is to get this right, I wonder if it's worth it. Particularly because all this change would do is enable an `.IsNull` syntax where `== default` would already do the same. Why is this so important?

Issue #596, opened 2022-06-24
Consider implementing Dispose pattern for handle typedefs


AArnott:
When a typedef struct represents a releasable handle, we could implement `IDisposable` so that C# `using` syntax would allow deterministic release of these handle.

elachlan:
I think this will be blocked by the solution in #610, which should establish how to reliably identify what nativedef structs are handles.

AArnott:
I don't think so... #610 is about picking which structs should get an `IsNull` property, but _this_ issue only needs to know whether a given typedef struct has a releaser method, which we already know.

elachlan:
Yes, I guess we have to know HOW to release it, to be able to release it. Adding this won't be too bad.

elachlan:
Should this be able to be turned on/off by a setting? Or would there be no benefit in that?

AArnott:
I don't anticipate any downside to it, so I wouldn't add a switch unless a downside presents itself.

elachlan:
@AArnott if we add `IDisposable` to a struct, do we also need to add an `bool ownsHandle` similar to `safehandles`? That way we can handle when a method is tagged with `DoNotReleaseAttribute`?

jnm2:
The handle struct would no longer be blittable and would also be subject to tearing.

elachlan:
So we can't use `IDisposable` then on any struct that is used by a method which has the `DoNotReleaseAttribute`.

AArnott:
No, it's not the `IDisposable` that would present the problems that @jnm2 calls out. It's the additional field where we would store the `ownsHandle` value.
If folks want .NET/C# to help with disposal of native resources, SafeHandle is the standard. The typedef structs are the native primitive that folks use when they specifically want to control lifetime themselves and/or cannot take the GC pressure of a SafeHandle. To make the struct disposable may be interesting, but it blurs the line. And if it led to analyzers producing warnings when they are not disposed of, that would violate a major point of falling back to them in the first place.

So maybe we could do the IDisposable struct thing if we new analyzers don't and would never call out disposing them. But I'm leaning against doing it at all given that's SafeHandle territory.

elachlan:
That is probably a good idea. So should we close this issue or limit it to structs without associated methods tagged with `DoNotReleaseAttribute`? I don't really have an opinion either way.

AArnott:
[...]

PR #657, opened 2022-08-17
Adding IDisposable to typedefs

elachlan:
Fixes: #596

Copied and modified code from safehandles section.

**WIP:**
-  I will need to check for invalid handle values.
- The casting seems a bit strange and is causing test failures. 
- Check if struct is used by a method, which has the `DoNotReleaseAttribute`. Struct is not disposable if this is the case.
- Have not worked on any of the tests yet


AArnott:
We should usually agree on a design on the issue (including working on some of the issues you call out like when they should be disposable and not) before a PR is created to avoid wasted effort on a design or idea that won't move forward.

PR #666, opened 2022-08-24
Improvements to HRESULT

### LINKED ISSUE #664: Improvements to `HRESULT`
Add `ToString` to print 0x hex syntax and `HRESULT_FROM_WIN32` to `HRESULT` to improve usability.

elachlan:
This overrides `ToString()` to produce the familiar `0x80001234` format for an HRESULT.
It also adds the `HRESULT_FROM_WIN32` macro.

Fixes #664


elachlan:
I am unsure what to do to fix the build error. Do I need to specify a dependency somewhere?

src/Microsoft.Windows.CsWin32/templates/HRESULT.cs
```diff
@@ -61,7 +61,9 @@ internal HRESULT ThrowOnFailure(IntPtr errorInfo = default)
         return this;
     }
 
-    public override string ToString() => this.Value.ToString();
+	internal static HRESULT HRESULT_FROM_WIN32(WIN32_ERROR error) => new(((int)error & 0x0000FFFF) | unchecked((int)0x80070000));
 
-    internal string ToString(string format, IFormatProvider formatProvider) => ((uint)this.Value).ToString(format, formatProvider);
+	public override string ToString() => this.Value.ToString("X");
```
AArnott, line 66: We want it in this format: `0x00005678`. That is, a `0x` prefix and padded so it ends with 8 hex characters.

src/Microsoft.Windows.CsWin32/templates/HRESULT.cs
```diff
@@ -61,7 +61,9 @@ internal HRESULT ThrowOnFailure(IntPtr errorInfo = default)
         return this;
     }
 
-    public override string ToString() => this.Value.ToString();
+	internal static HRESULT HRESULT_FROM_WIN32(WIN32_ERROR error) => new(((int)error & 0x0000FFFF) | unchecked((int)0x80070000));
```
AArnott, line 64: Are we sure the conversion is this simple? I'm looking at [this doc](https://docs.microsoft.com/en-us/windows/win32/api/winerror/nf-winerror-hresult_from_win32#remarks) and it seems more complicated than this.

[This C# HResult struct](https://github.com/dotnet/pinvoke/blob/main/src/Windows.Core/HResult.cs) may be a resource as well.

elachlan: Should I be adding via generator is the dependency is added when importing HRESULT?

elachlan: @AArnot, I guessing winforms did what they did because `(FACILITY_WIN32 << 16) | 0x80000000)` effectively is `unchecked((int)0x80070000)`? I don't know for sure, but I think its the bitshift on the const?

AArnott: > Should I be adding via generator so the dependency is added when importing HRESULT?

Yes, we absolutely should ensure the referenced types are generated. We can probably combine the template simplicity with a special step in the generator that 'knows' to add the types referenced in the template to try to capture the best of both worlds.

The difference in the expression that I was focusing on is that negative WIN32_ERROR is supposed to be taken as-is. Only positive values get this special bit-wise treatment.

elachlan: > Yes, we absolutely should ensure the referenced types are generated. We can probably combine the template simplicity with a special step in the generator that 'knows' to add the types referenced in the template to try to capture the best of both worlds.

I have no idea how to implement that.

> The difference in the expression that I was focusing on is that negative WIN32_ERROR is supposed to be taken as-is. Only positive values get this special bit-wise treatment.

The negative test was added in windows xp. Nothing on why though. I imagine the negative values were added around then and maybe they were managed better in `WIN32_ERROR`? Who knows.
https://matthewvaneerde.wordpress.com/2007/12/28/the-evolution-of-hresult_from_win32/

Raymond Chen also wrote an article on doing the reverse mapping which might be useful.
https://devblogs.microsoft.com/oldnewthing/20061103-07/?p=29133

elachlan: @russkie sorry to bother you. But do you know why there is a negative check in `HRESULT_FROM_WIN32` and why there isn't a bit shift on it?

AArnott: Regardless of the reason the negativity check exists, for auditing purposes I'd prefer that we kept as close to the same syntax as the header files as possible. 
I can help with the testing.

elachlan: Is there anywhere else in the code where we detect dependencies and then include them?

RussKie: @ecalalang I am afraid I don't know. I think you know about this speific area way more than me at this point :)

elachlan: Thanks, do you know someone that might know?

RussKie: May be try nerd sniping @oldnewthing? :D 

My first point of contact for such things is generally @JeremyKuhne, but he's OOF right now.

JeremyKuhne: @elachlan, @RussKie presumably the negative check there is to deal with things that are already `HRESULT`s or possibly filter out other ["custom" codes](https://docs.microsoft.com/windows/win32/debug/system-error-codes#third-party-error-codes) that might be set on last error, but I'm not 100% sure. In any case, something that was negative wouldn't make much sense wrapping in an `HRESULT` and claiming it was a Win32 error. I'd just follow along with the actual implementation.

AArnott:
I refactored the `HRESULT_FROM_WIN32` into its own optional 'macro' (a new concept) so that `WIN32_ERROR` doesn't always have to be (exhaustively) generated every time `HRESULT` is required.

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -477,34 +492,35 @@ private IEnumerable<MemberDeclarationSyntax> NamespaceMembers
...
-            for (int i = 0; i < members.Count(); i++)
+            int i = 0;
+            foreach (IGrouping<string, MemberDeclarationSyntax> entry in members)
             {
-                IGrouping<string, MemberDeclarationSyntax> entry = members.ElementAt(i);
-                if (i == 0)
+                ClassDeclarationSyntax partialClass = DeclarePInvokeClass(entry.Key)
+                    .AddMembers(entry.ToArray())
+                    .WithLeadingTrivia(ParseLeadingTrivia(string.Format(CultureInfo.InvariantCulture, PartialPInvokeContentComment, entry.Key)));
+                if (i++ == 0)
```
RussKie, line 501: [nit] May be we could avoid combining two operations as it could confuse some readers, and may inadvertenly introduce bugs if this code is modified/extended later.

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -1038,6 +1086,34 @@ public bool TryGenerateConstant(string possiblyQualifiedName, out IReadOnlyList<
...
+    /// </summary>
+    /// <param name="macroName">The name of the macro. Never qualified with a namespace.</param>
+    /// <param name="preciseApi">Receives the canonical API names that <paramref name="macroName"/> matched on.</param>
+    /// <returns><see langword="true"/> if a match was found and the macro generated; otherwise <see langword="false"/>.</returns>
+    public bool TryGenerateMacro(string macroName, out IReadOnlyList<string> preciseApi)
+    {
+        if (macroName is null)
+        {
+            throw new ArgumentNullException(nameof(macroName));
+        }
```
RussKie, line 1100: If we're targeting .NET 6+, this could be changed to [`ArgumentNullException.ThrowIfNull`](https://docs.microsoft.com/dotnet/api/system.argumentnullexception.throwifnull)

src/Microsoft.Windows.CsWin32/templates/PInvokeClassMacros.cs
```diff
@@ -0,0 +1,15 @@
+﻿/// <summary>
+/// This class wrapper is stripped so that individual macros can be requested for generation.
+/// </summary>
+internal class PInvokeClassMacros
+{
+	/// <summary>
```
RussKie, line 6: Not sure what the overall policy in this repo, but this file is using tabs, whereas other files appear to be using spaces. Perhaps something is amiss in .editorconfig?

src/Microsoft.Windows.CsWin32/Generator.cs
AArnott: I didn't know about that. Cool. But this generator targets netstandard2.0 so the source generator can run in C# compiler contexts that are hosted by .NET Framework.

elachlan: Maybe we can write our own helper methods?

RussKie: Ah right, no worries then.

In the designer repo we added some throw helpers with #pragmas, something _could_ be useful here, e.g.:
```cs
    /// <summary>
    ///  If <paramref name="argument"/> is <see langword="null"/>, throw an <see cref="ArgumentNullException"/>.
    ///  Otherwise, return the value of <paramref name="argument"/>.
    /// </summary>
    /// <exception cref="ArgumentNullException"/>
    public static T OrThrowIfArgumentIsNull<T>(
        [NotNull] this T? argument,
        [CallerArgumentExpression("argument")] string? paramName = null)
    {
#if NET6_0_OR_GREATER
        ArgumentNullException.ThrowIfNull(argument, paramName);
#else
        ThrowIfArgumentIsNull(argument, paramName);
#endif

        return argument;
    }

    /// <summary>
    ///  If <paramref name="argument"/> is <see langword="null"/>, throw an <see cref="ArgumentNullException"/>.
    /// </summary>
    /// <exception cref="ArgumentNullException"/>
#if NET6_0_OR_GREATER
    [Obsolete("Use ArgumentNullException.ThrowIfNull(...) instead.")]
#endif
    public static void ThrowIfArgumentIsNull(
        [NotNull] object? argument,
        [CallerArgumentExpression("argument")] string? paramName = null)
    {
#if NET6_0_OR_GREATER
        ArgumentNullException.ThrowIfNull(argument, paramName);
#else
        if (argument is null)
        {
            ThrowHelper.ThrowArgumentNullException(paramName);
        }
#endif
```

src/Microsoft.Windows.CsWin32/templates/PInvokeClassMacros.cs
AArnott: Oh wait, I guess there's an .editorconfig file in the templates directory itself that prescribes tabs. Now that I recall, all the code we generate uses tabs, so we want the templates to use tabs as well.

RussKie: That makes sense. I commented because it looked inconsistent :)

AArnott: I'll create a PR to use tabs consistently in the templated files that are using spaces.

src/Microsoft.Windows.CsWin32/Generator.cs
AArnott: I've pushed [a change](5632b2d840ebe5ee698379088f20bf4c333dd4de) to apply this.

src/Microsoft.Windows.CsWin32/templates/PInvokeClassMacros.cs
AArnott: I pushed as ee1ade7b343f97f210d2fb90262add61a6063a4b

src/Microsoft.Windows.CsWin32/Generator.cs
RussKie: 🚀

src/Microsoft.Windows.CsWin32/templates/PInvokeClassMacros.cs
RussKie: 🚀

src/Microsoft.Windows.CsWin32/Generator.cs
AArnott: We already have a bunch of helper methods in a `Require` class (in the Microsoft.VisualStudio.Validation package) that we use for this purpose, but as this a source generator and dependencies have a tendency to create new issues, I have to do without them.
While we could re-introduce the helper methods as source code within the project itself, I don't think it's worth it. We hardly have null checks anyway because we use nullable ref annotations everywhere anyway. I only included it here because it was in a public method.
IMO if it took until .NET 6 to add a helper method to the runtime, we can get by with a manual `if` check in a few places. :)

elachlan: You should check out the other performance enhancements in .NET7, its a chapter book! https://devblogs.microsoft.com/dotnet/performance_improvements_in_net_7

PR #672, opened 2022-08-26
Add DateTime operators to SYSTEMTIME

### LINKED ISSUE #667: Add `DateTime` operators to `SYSTEMTIME`
Issue came up in a PR(https://github.com/dotnet/winforms/pull/7662) as apart of the winforms conversion effort.

https://github.com/dotnet/winforms/blob/74e08305a520abc8cf176350072784c6d8d98ddc/src/System.Windows.Forms.Primitives/src/Windows/Win32/Foundation/SYSTEMTIME.cs

https://github.com/dotnet/winforms/issues/7468

elachlan:
Fixes #667 and #673

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -3726,6 +3726,7 @@ public class Generator : IDisposable
         {
             case "RECT":
             case "SIZE":
+            case "SYSTEMTIME":
                 members.AddRange(this.ExtractMembersFromTemplate(name.Identifier.ValueText));
                 break;
             default:
@@ -6315,6 +6316,28 @@ public class Generator : IDisposable
 
         public override SyntaxNode? VisitExpressionStatement(ExpressionStatementSyntax node) => base.VisitExpressionStatement(this.WithIndentingTrivia(node));
 
+        public override SyntaxNode? VisitInitializerExpression(InitializerExpressionSyntax node)
+        {
+            node = node
+                .WithOpenBraceToken(Token(TriviaList(this.IndentTrivia), SyntaxKind.OpenBraceToken, TriviaList(LineFeed)))
+                .WithCloseBraceToken(Token(TriviaList(this.IndentTrivia), SyntaxKind.CloseBraceToken, TriviaList(LineFeed)))
+                .WithTrailingTrivia(TriviaList());
+            using var indent = new Indent(this);
+            return base.VisitInitializerExpression(node);
+        }
+
+        public override SyntaxNode? VisitAssignmentExpression(AssignmentExpressionSyntax node)
+        {
+            if (node.Parent is InitializerExpressionSyntax)
+            {
+                return base.VisitAssignmentExpression(this.WithIndentingTrivia(node));
+            }
+            else
+            {
+                return base.VisitAssignmentExpression(node);
+            }
+        }
+
         public override SyntaxNode? VisitTryStatement(TryStatementSyntax node) => base.VisitTryStatement(this.WithIndentingTrivia(node));
 
         public override SyntaxNode? VisitFinallyClause(FinallyClauseSyntax node) => base.VisitFinallyClause(this.WithIndentingTrivia(node));
@@ -6389,7 +6412,10 @@ public class Generator : IDisposable
             }
         }
 
-        public override SyntaxNode? VisitReturnStatement(ReturnStatementSyntax node) => base.VisitReturnStatement(node.WithLeadingTrivia(this.IndentTrivia));
+        public override SyntaxNode? VisitReturnStatement(ReturnStatementSyntax node)
+        {
+            return base.VisitReturnStatement(this.WithIndentingTrivia(node));
+        }
 
         public override SyntaxToken VisitToken(SyntaxToken token)
         {
--- a/src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
+++ b/src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
@@ -1,0 +1,42 @@
+﻿partial struct SYSTEMTIME
+	: IEquatable<SYSTEMTIME>
+{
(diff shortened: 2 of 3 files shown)


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -3726,6 +3726,7 @@ private StructDeclarationSyntax DeclareStruct(TypeDefinitionHandle typeDefHandle
         {
             case "RECT":
             case "SIZE":
+            case "SYSTEMTIME":
```
AArnott, line 3729: Given this now has special behavior, can you please add a test that generates only `SYSTEMTIME` and ensures no compiler diagnostics are produced? There are several similar tests already in existence for you to copy. In fact if we don't already have a `[Theory]` for all these templated APIs, could you add one theory method for all these templated APIs?

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
```diff
@@ -0,0 +1,28 @@
+﻿partial struct SYSTEMTIME
+{
+	public static implicit operator global::System.DateTime(SYSTEMTIME sysTime)
+	{
+		if (sysTime.wYear <= 0 || sysTime.wMonth <= 0 || sysTime.wDay <= 0)
+		{
+			Debug.Fail("Incorrect SYSTEMTIME info!");
+			return global::System.DateTime.MinValue;
+		}
+
```
AArnott, line 10: I don't think this is acceptable to just return `DateTime.MinValue` in an error condition. For WinForms they're free to swallow errors if they want, but we can't do that in a general system like this.

An interesting scenario that this check includes (among others) is `default(SYSTEMTIME)`, which evidently cannot be converted. Maybe we could do something special where we say `if (sysTime == default) return default;` in both conversion operators. Then at least structs with all zeros can convert between each other with no information loss and no exception. At that point, exceptions only get thrown when there is actually invalid data set into the struct.
Thoughts?

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
```diff
@@ -0,0 +1,28 @@
...
+	public static implicit operator SYSTEMTIME(global::System.DateTime time) => new SYSTEMTIME
+	{
+		wYear = (ushort)time.Year,
+		wMonth = (ushort)time.Month,
+		wDayOfWeek = (ushort)time.DayOfWeek,
+		wDay = (ushort)time.Day,
+		wHour = (ushort)time.Hour,
+		wMinute = (ushort)time.Minute,
+		wSecond = (ushort)time.Second,
+		wMilliseconds = (ushort)time.Millisecond
```
AArnott, line 26: All these should be surrounded with a `checked { }` block since we're truncating values. While we may not expect *valid* values to ever exceed `ushort.MaxValue`, this code shouldn't be the means of silently truncating invalid values into something that may appear to be more acceptable.

AArnott (changes requested):
Thanks for getting this started.

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
AArnott: As an aside, given this will be different behavior from what WinForms was used to, we should call out to them when they adopt this that it's different, and they should carefully review where they rely on their old implicit conversion operator to avoid regressions. 
This BTW is an argument for not declaring implicit operators except where they can be trivially implemented. I'm still on the fence as to whether these should be implicit or explicit. When explicit, it would be far easier for WinForms to audit their code (because errors will be everywhere they used to use their implicit ones) and either allow use of an explicit operator or call into a method they define that retains their old proprietary behavior.

elachlan: I think I have implemented what you asked. Did we want to keep the default and the year/month/day comparison separate and throw an exception when those values are invalid?

I think they should be explicit. SYSTEMTIME for year "The valid values for this member are 1601 through 30827."
https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-systemtime

For DateTime "The year, between 1 and 9999."
https://docs.microsoft.com/en-us/dotnet/api/system.datetime.year?view=net-6.0

We **will** get data loss when converting between them.

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
```diff
@@ -0,0 +1,75 @@
...
+			return default;
+		}
+
+		// DateTime gets DayOfWeek automatically
+		return new global::System.DateTime(sysTime.wYear,
+			sysTime.wMonth, sysTime.wDay, sysTime.wHour,
+			sysTime.wMinute, sysTime.wSecond, sysTime.wMilliseconds);
+	}
+
+	public static explicit operator SYSTEMTIME(global::System.DateTime time)
```
elachlan, line 56: This could maybe be implicit. But if it is ever used in interop it might cause issues because the minimum value for year is `1601`.
https://docs.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-systemtime

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
AArnott: > We will get data loss when converting between them.

If data loss is inevitable, they should *not* be `implicit` conversion operators.
If we work for all arguably valid data, and can throw in the unconvertible cases, that *may* qualify for implicit operators.
Otherwise, explicit operators would probably be better.

AArnott:
@elachlan I'm on jury duty last week and next week, FYI. So my participation will be weak and delayed through the end of that.

src/Microsoft.Windows.CsWin32/templates/SYSTEMTIME.cs
AArnott: Let's leave it as explicit for now. It's safer to move from explicit to implicit later than the other way around.

AArnott (approved):

AArnott:
Thanks, @elachlan

PR #679, opened 2022-09-01
Update metadata to 31.0.4-preview

AArnott:


AArnott:
FullGeneration is failing due to https://github.com/microsoft/win32metadata/issues/1207

PR #735, opened 2022-10-18
Add `CharSet = Unicode` to `extern` methods with `char` parameters

### LINKED ISSUE #734: `VkKeyScanW`/`VkKeyScanExW` are missing `CharSet = CharSet.Unicode`
These functions and probably a lot more we don't know of should have `CharSet = CharSet.Unicode`.
- https://learn.microsoft.com/en-au/windows/win32/api/winuser/nf-winuser-vkkeyscanw
- https://learn.microsoft.com/en-au/windows/win32/api/winuser/nf-winuser-vkkeyscanexw

We should probably see if we can extract the information somehow, or if that isn't possible manually tag them.

`WinUser.h`:
https://github.com/microsoft/win32metadata/blob/f86785bec72eef8aa9c9cf5a84fc6d446abe2db5/generation/WinSDK/RecompiledIdlHeaders/um/WinUser.h#L5989-L6003

Related: https://github.com/dotnet/winforms/issues/7945
Winforms Tracking: https://github.com/dotnet/winforms/issues/7468

AArnott:
Fixes #734

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -2300,7 +2300,7 @@ public class Generator : IDisposable
                 IdentifierName(Enum.GetName(typeof(ComInterfaceType), interfaceType)!))));
     }
 
-    private static AttributeSyntax DllImport(MethodImport import, string moduleName, string? entrypoint)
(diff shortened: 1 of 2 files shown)


elachlan:
Awesome, thank you!

PR #739, opened 2022-10-26
Emit COM interfaces with properties where possible

### LINKED ISSUE #738: Project COM getters and setters as properties
Consider APIs like Windows.UI.Accessibility.IUIAutomationElement.get_CachedAcceleratorKey. This should be exposed to .NET developers as a _property_ instead of a method. The `get_` prefix should be omitted.

I want this so I can use cswin32 to seamlessly replace interop assemblies generated by tlbimp. This change would also make cswin32 more consistent with the Rust win32 projection ([which also omits the get_ prefix](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/Accessibility/struct.IUIAutomationElement.html#method.CachedAccessKey)) and more consistent with CsWinRT.

The metadata already puts the SpecialName attribute on methods which need this treatment.

AArnott:
Closes #738


mcooley:
Very cool! Thank you!

PR #762, opened 2022-11-11
Observe `DangerousAddRef` result before proceeding.

### LINKED ISSUE #754: SafeHandle to typedef conversion should not ignore AddRef failure
### Discussed in https://github.com/microsoft/CsWin32/discussions/753

<div type='discussions-op-text'>

<sup>Originally posted by **User1785604260** November  7, 2022</sup>
I needed to use DangerousGetHandle in some of my own code, and wanted to use the generated CsWin32 code as a reference to see if I was doing so correctly. In looking at the generated CsWin32 code, I'm now wondering if its DangerousAddRef/GetHandle usage is correct. I am 1000% not an expert on this, so I may be completely incorrect.

What I noticed is that the success of DangerousAddRef is only used to determine whether to call DangerousRelease. If DangerousAddRef fails, the generated code still proceeds to call DangerousGetHandle and use that handle. From what I can tell this is incorrect, as failure means that the handle is closed if I'm understanding correctly.

If failure shouldn't be ignored, I'm not sure what the correct behavior would be instead. Maybe throwing ObjectDisposedException?

For example, here's an arbitrary method I picked that uses DangerousGetHandle:
````csharp
public static unsafe int DestroyPhysicalMonitor(SafeHandle hMonitor)
{
    bool hMonitorAddRef = false;
    try
    {
        winmdroot.Foundation.HANDLE hMonitorLocal;
        if (hMonitor is object)
        {
            hMonitor.DangerousAddRef(ref hMonitorAddRef);
            hMonitorLocal = (winmdroot.Foundation.HANDLE)hMonitor.DangerousGetHandle();
        }
        else
            hMonitorLocal = default(winmdroot.Foundation.HANDLE);
        int __result = PInvoke.DestroyPhysicalMonitor(hMonitorLocal);
        return __result;
    }
    finally
    {
        if (hMonitorAddRef)
            hMonitor.DangerousRelease();
    }
}
````

On a more advanced note, the [documentation for DangerousAddRef](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle.dangerousaddref?view=net-6.0#remarks) also indicates that a Constrained Execution Region should be used. I've never touched those and don't know what proper usage of that would look like, and if it's relevant to CsWin32.</div>

Thank you @User1785604260 for bringing this up.

AArnott:
Fixes #754

### DIFF
--- a/src/Microsoft.Windows.CsWin32/FastSyntaxFactory.cs
+++ b/src/Microsoft.Windows.CsWin32/FastSyntaxFactory.cs
@@ -135,6 +135,10 @@ internal static class FastSyntaxFactory
 
     internal static ThrowStatementSyntax ThrowStatement(ExpressionSyntax expression) => SyntaxFactory.ThrowStatement(Token(TriviaList(), SyntaxKind.ThrowKeyword, TriviaList(Space)), expression, Semicolon);
 
+    internal static ThrowExpressionSyntax ThrowExpression(ExpressionSyntax expression) => SyntaxFactory.ThrowExpression(Token(TriviaList(), SyntaxKind.ThrowKeyword, TriviaList(Space)), expression);
+
+    internal static ExpressionSyntax NameOfExpression(IdentifierNameSyntax identifierName) => SyntaxFactory.InvocationExpression(IdentifierName("nameof"), ArgumentList(SingletonSeparatedList(Argument(identifierName))));
+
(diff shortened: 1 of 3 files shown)


AaronRobinsonMSFT:
This isn't necessary. Sorry for the slow reply. The `DangerousAddRef` will throw on failure to add ref.

AArnott:
Thanks, @AaronRobinsonMSFT. I'll revert this change.

PR #767, opened 2022-11-11
Emit a stable reference to the interface guid

AArnott:
This way, the reference can be converted by others into a pointer and passed around without fear that the data will move.

The new property looks like this:

```cs
static ref readonly Guid IComIID.Guid
{
    get
    {
        ReadOnlySpan<byte> data = new byte[] {
            0x0B,0x01,0x00,0x00,0x00,0x00,0x00,0x00,0xC0,0x00,0x00,0x00,0x00,0x00,0x00,0x46 };
        return ref Unsafe.As<byte, Guid>(ref MemoryMarshal.GetReference(data));
    }
}
```

Also add non-marshaling runtime test project so we can verify the emitted guid has the right value.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/sdk:6.0.401-focal
+FROM mcr.microsoft.com/dotnet/sdk:7.0.100-jammy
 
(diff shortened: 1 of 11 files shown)


AArnott:
CC: @tannergooding @JeremyKuhne

JeremyKuhne (approved):
🚀 Nice!

Issue #774, opened 2022-11-15
[PreserveSig] inconsistently applied in IPicture


JeremyKuhne:
## Actual behavior

When generating `IPicture` we get the following errors in the latest build:

```
Windows.Win32.IPicture.g.cs(87,36): error CS0082: Type 'IPicture' already reserves a member called 'set_hPal' with the same parameter types
Windows.Win32.IPicture.g.cs(379,37): error CS0082: Type 'IPicture.Interface' already reserves a member called 'set_hPal' with the same parameter types
```

It's generating the following:

``` C#
winmdroot.System.Ole.OLE_HANDLE hPal
{
    get
    {
        fixed (IPicture* pThis = &this)
        {
            winmdroot.System.Ole.OLE_HANDLE __result;
            ((delegate *unmanaged [Stdcall]<IPicture*,winmdroot.System.Ole.OLE_HANDLE* ,winmdroot.Foundation.HRESULT>)lpVtbl[4])(pThis, &__result).ThrowOnFailure();
            return __result;
        }
    }
}
```

It isn't doing `[PreserveSig]`, while the setter is. Maybe something to do with the introduction of `OLE_HANDLE`?

## Expected behavior

All methods are `[PreserveSig]` when `"allowMarshalling"` is `false`.

## Repro steps

1. `NativeMethods.txt` content:
```
IPicture
```

2. `NativeMethods.json` content:
```json
{
  "$schema": "https://aka.ms/CsWin32.schema.json",
  "allowMarshaling": false,
  "useSafeHandles": false
}
```

### Context

- CsWin32 version: 0.2.123-beta
- Target Framework: net8.0

AArnott:
I can repro with a test. It's probably a regression from #738.

AArnott:
The compilation break is due to https://github.com/microsoft/win32metadata/issues/1367. I still don't understand how this bug is caught by a new, special purpose test but not caught in our catch-all full generation tests.

AArnott:
> All methods are `[PreserveSig]` when "allowMarshalling" is false.

Hmmm... so you don't want specialname `get_` and `set_` methods to be declared as properties on structs?

AArnott:
Our 'full generation' test apparently skips all COM interfaces. 😲

JeremyKuhne:
> Hmmm... so you don't want specialname get_ and set_ methods to be declared as properties on structs?

The question is what you do with HRESULT codes. As long as we have both so we can check them manually, I'm fine.

[...]

PR #776, opened 2022-11-15
Add IPicture test and fix the failure

AArnott:
Fixes the non-compilable generated code reported in #774.
This change does _not_ turn off property declarations in COM structs. That (if necessary) can come in a follow-up change.


JeremyKuhne (approved):
👍🏼

PR #777, opened 2022-11-15
Fix full generation tests

AArnott:
Our full generation testing was actually omitting many COM interfaces. Closing that hole revealed a few bugs where we can emit C# that will not compile.
These changes address the test hole and the issues that closing it exposed.


AArnott:
There is a test failure in generating `IAudioProcessingObjectConfiguration` due to https://github.com/microsoft/win32metadata/issues/1368.

PR #848, opened 2023-01-17
Null-coalescing in friendly overload

### LINKED ISSUE #277: Codegen for nullable structs in friendly overloads could be optimized
Example from `CredUIPromptForWindowsCredentials`:

```cs
CREDUI_INFOW pUiInfoLocal = pUiInfo.HasValue ? pUiInfo.Value : default(CREDUI_INFOW);
uint __result = PInvoke.CredUIPromptForWindowsCredentials(pUiInfo.HasValue ? &pUiInfoLocal : null, ...);
```

`pUiInfo.HasValue ? pUiInfo.Value : default(CREDUI_INFOW);` could be replaced with `pUiInfo ?? default` which is more idiomatic C# and which compiles to `pUiInfo.GetValueOrDefault()` which IIRC is jitted as direct field access. Whereas calling `.Value` adds an additional redundant check and looks more complex.

(Version used: 0.1.422-beta)

davidegiacometti:
Fixes #277
Use null-coalescing in friendly overload.
Generate `x ?? default(TYPE)` instead of `x.HasValue ? x.Value : default(TYPE)`.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
@@ -329,10 +329,8 @@ public partial class Generator
                     leadingStatements.Add(
                         LocalDeclarationStatement(VariableDeclaration(elementType)
                             .AddVariables(VariableDeclarator(localName.Identifier).WithInitializer(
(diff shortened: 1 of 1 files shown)


AArnott (approved):
Love it. Thanks.

PR #890, opened 2023-03-18
add net35 fix

peaceshi:


peaceshi:
@microsoft-github-policy-service agree

AArnott:
This could work. Thank you for offering the code change. Looking more closely at this, it appears the break is because overloads aren't created that net35 is fully compatible with, and it's not yet clear to me why these overloads aren't being created. If we can fix that, we won't need to duplicate the definition of this method by hand.

Also, just as a nit, the indentation isn't consistent in the two regions. But we needn't fix this yet while I look to see if there's a better fix.

AArnott:
Thank you again for your contribution. I found a more general fix that addresses the original bug, and much more. 

Check out #892.

PR #930, opened 2023-05-08
Leverage AlsoUsableForAttribute for more interoperable typedefs

### LINKED ISSUE #199: Leverage `AlsoUsableForAttribute` for more interoperable typedefs

### LINKED ISSUE #924: HINSTANCE is missing
## Actual behavior

IUIFramework.LoadUI() first parameter "instance" now has the type "HMODULE".
CsWin32 version 0.2.223-beta had for the first parameter the type "HINSTANCE".

## Expected behavior

The [documentation](https://learn.microsoft.com/en-us/windows/win32/api/uiribbon/nf-uiribbon-iuiframework-loadui) and also the header file UiRibbon.h say that the first parameter type is a HINSTANCE

## Repro steps
```
IUIFramework Framework;
private void Method(HINSTANCE hInstance, string resourceIdent)
{
    Framework = new UIRibbonFramework() as IUIFramework;
    Framework.LoadUI(hInstance, resourceIdent);
}
```

```
UIRibbonFramework
IUIFramework

```

```json
```

### Context

- CsWin32 version: [0.2.229-beta]
- Target Framework: [e.g. `netstandard2.0`]

mikebattista:
Fixed #199 
Fixed #924 

Adds implicit conversion operators between AlsoUsableFor types. An existing example included in the tests is `BCRYPT_KEY_HANDLE` -> `BCRYPT_HANDLE`.

When https://github.com/microsoft/win32metadata/issues/1517 is fixed, this will also support `HMODULE <=> HINSTANCE` conversions.


src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
```diff
@@ -319,6 +338,41 @@ private IEnumerable<MemberDeclarationSyntax> CreateCommonTypeDefMembers(Identifi
...
+        }
+        else
+        {
+            throw new GenerationFailedException("Unexpected handle type.");
+        }
+
+        List<string> alsoUsableForValues = new();
+        QualifiedTypeDefinition td = tdh.Resolve();
+        foreach (CustomAttributeHandle ah in td.Definition.GetCustomAttributes())
+        {
```
AArnott, line 363: Actually since evidently what we want is the `TypeDefinition` and our caller already has it, we should just pass that in directly rather than pass in any handle at all and require this method to re-look it up.

AArnott (changes requested):
It looks good. Just a couple changes please.

src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
mikebattista: I'm not sure what the equivalent calls are later on if we pass in just a TypeDefinition. Could you fix this one?

mikebattista: I found `this.Reader`. Is that the equivalent of `td.Reader`? It seems to work.

AArnott (approved):
Great. Thanks. I don't think I ever finished explaining the set of tests we have to run locally. I'll run them myself and then merge.

PR #1017, opened 2023-08-11
Implement `SetLastError` ourselves on .NET when marshaling is not allowed

### LINKED ISSUE #600: Implement SetLastError ourselves when allowMarshaling is false
It's worth noting that CsWin32's support for disabling marshalling won't actually be usable with [`DisableRuntimeMarshallingAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.disableruntimemarshallingattribute) yet because of `SetLastError`: https://docs.microsoft.com/en-us/dotnet/standard/native-interop/disabled-marshalling#disabled-features

_Originally posted by @alexrp in https://github.com/microsoft/CsWin32/issues/593#issuecomment-1179162557_

AArnott:
Closes #600

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.Extern.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Extern.cs
@@ -216,9 +216,13 @@ public partial class Generator
                 }
             }
 
+            bool setLastError = (import.Attributes & MethodImportAttributes.SetLastError) == MethodImportAttributes.SetLastError;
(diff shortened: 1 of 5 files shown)


jlaanstra:
Thanks @AArnott!

jlaanstra:
Any chance we could get an updated package with this change?

AArnott:
You can get it now per these instructions: https://github.com/microsoft/CsWin32/issues/600#issuecomment-1692165318

We'll probably release a new version to nuget.org soon though.

PR #1020, opened 2023-08-11
Add ToString() methods to ordinary typedef structs

AArnott:


elachlan:
We are running into an issue where we have implemented `BOOL.ToString() => ((bool)this).ToString();` This clashes with this change. Obvious solution would be to rename the winforms `ToString` to something else. 

Does returning to hex value for `BOOL` make sense though?

AArnott:
I could imagine CsWin32 special casing BOOL to print a boolean value. However since some uses of BOOL stick more values into it than the customary two, it may indeed be useful to print its full value as an integer.

AArnott:
Another option is to notice that your compilation already defines a ToString method and suppress automatic generation.

PR #1114, opened 2024-01-12
Allow `ComHelpers` class to be non-partially declared by compilation

AArnott:


JeremyKuhne (approved):
I believe this is right? As long as it allows you to generate these if they are partially defined in the current assembly. (Which is how I'm reading this)

AArnott:
Correct. As with other types you might declare yourself, only _non_-partial declarations will turn off CsWin32 generation.

PR #1375, opened 2025-02-21
Bump System.Text.Encodings.Web to 4.7.2 (again)

AArnott:


AArnott:
This fails the VS 2019 integration test:

> ##[warning]CSC(0,0): Warning CS8785: Generator 'SourceGenerator' failed to generate source. It will not contribute to the output and compilation errors may occur as a result. Exception was of type 'FileNotFoundException' with message 'Could not load file or assembly 'System.Text.Encodings.Web, Version=4.0.5.0, Culture=neutral, PublicKey***=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.'

PR #1388, opened 2025-03-16
Update getting-started.md

LWChris:
Moved condition to the front of the sentence, so readers will immediately see whether to skip this step or not.


AArnott (approved):
Sounds good. Thanks.

LWChris:
Is there something I can or must do for that one pending check? It's not moving since when I opened the request.

LWChris:
@Aarnott Can you tell me what needs to be done for this to be merged?

PR #1396, opened 2025-04-09
Make `IComIID.Guid` public

### LINKED ISSUE #1394: Make IComIID.Guid public

I disabled marshalling and every COM definition inherits from `IComIID` but since `IComIID.Guid` doesn't have `public` modifer, I can't access it.

Can you add that modifier here?

https://github.com/microsoft/CsWin32/blob/10ff614acb54f6d5a5dbb77c59bd8786e5c99097/src/Microsoft.Windows.CsWin32/Generator.Com.cs#L867-L872

None

**Additional context**

Also, what's the reason why you added another Guid property (`IID_Guid`) here (plus the styling looks odd)?

Image

Lamparter:
- Closes #1394

Makes `[IComIID].Guid` public:
```diff
- static ref readonly Guid IComIID.Guid
+ public static ref readonly Guid IComIID.Guid
```

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.Com.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Com.cs
@@ -864,10 +864,10 @@ public partial class Generator
 
                 BlockSyntax getBody = Block(dataDecl, returnStatement);
 
-                // static ref readonly Guid IComIID.Guid { get { ... } }
+                // public static ref readonly Guid IComIID.Guid { get { ... } }
                 PropertyDeclarationSyntax guidProperty = PropertyDeclaration(IdentifierName(nameof(Guid)).WithTrailingTrivia(Space), ComIIDGuidPropertyName.Identifier)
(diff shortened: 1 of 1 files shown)


AArnott (changes requested):
Thank you for contributing.

You cannot combine `public` with explicit interface implementation syntax. This would produce a compiler error in emitted code. [sample](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQMwAJboMLpOgSWwHsBbAggEXQG9k1M4A2dAJwFMAzN9gQ2GIA7ADYBPdAHEArgEtghEuSoA6aXPQBeAHzoALgAtWxAO7pB7UwFEAHgGN2AB10yhACgCUAbmQBfZMhlBXXZWTl57BTIKajokdHjGFl4AIwBnXVZw3R5uDn4hMUlZeTV5GnQAc3ZdT3Q/JB8gA==).

I'm closing the PR and responding to the original issue.

PR #1405, opened 2025-05-01
ArrayPool can be larger than requested resulting in freeing uninitialized GCHandles

jlaanstra:
Code generator assumes the ArrayPool size is exact but in reality it can be larger, resulting in freeing uninitialized GCHandles which throws.


anna-is-cute:
What's currently blocking this PR from being merged?

Our error logs are full of handle not initialised errors due to this, and we had to downgrade to 0.3.162 for the time being, so it would be nice for this to get fixed!

bricelam:
+1 We've been stuck on 0.3.162 for seven months now. This PR has gone five months without comment from the maintainers.

@copilot can you help with this one? :trollface:

jevansaks:
I'm taking a look at this PR now. @jlaanstra we always include a test case for changes like this. I'll add one to your branch and also merge in latest main. Stay tuned.

PR #1445, opened 2025-07-24
Add custom DebuggerDisplay for HRESULT

misaz:
PR adds support for using customized DebuggerDisplay for customized types (templates) instead of hardocoding `DebuggerDisplay` to structure's first field name. It is designed to work in a way, that when there is property named `DebuggerDisplay`, it is used instead. Actually there already were some custom DebuggerDisplays in templates, but these only work with [`SpecialTypeDefNames`](https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/Generator.Invariants.cs#L63) because for normal types, only Members are obtained and rest of the template is thrown away and recrated including [DebuggerDisplay attribute](https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs#L247).

PR also uses newly introduced capability and adds custom `DebuggerDisplay` for HRESULT type. Custom visualizer show HEX error code number followed by humman readable description of HRESULT error code:

obrazek

It allows developer debugging apps to understand errors without need to use Error Lookup tool or searching online.

Message is internally obtained using avalaible .NET APIs (two variants are implemented depending on .NET version, both should produce same results because internally both end up using `FormatMessageW`). As you can see, FormatMessage provides localized message. I use Czech windows, so message is in Czech. For debugging it should be no problem.

It also brings debugger visualization experience on par with native debugger which has similar natvis (or maybe hardcoded, who knows?) visualizer for HRESULT as well.

I was also considering adding similar for NTSTATUS, but obtaining message for NTSTATUSes is not that easy because it requires passing NTDLL module handle (+ one more flag) to FormatMessageW and this is not exposed to .NET API. But it can be done manually. Instead of bloating NTSTATUS template in horroble way, I decided to step back on it. Also note, that native debugger do not support visualizer for NTSTATUS as well.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
@@ -239,12 +239,22 @@ public partial class Generator
             structModifiers = structModifiers.Add(TokenWithSpace(SyntaxKind.UnsafeKeyword));
         }
 
+        string debuggerDisplay;
+        if (members.Where(x => x is PropertyDeclarationSyntax && ((PropertyDeclarationSyntax)x).Identifier.ValueText == "DebuggerDisplay").Any())
+        {
+            debuggerDisplay = "{DebuggerDisplay,nq}";
+        }
+        else
+        {
+            debuggerDisplay = "{" + fieldName + "}";
+        }
+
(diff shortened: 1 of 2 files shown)


misaz:
@microsoft-github-policy-service agree

AArnott (approved):
Thanks! This is awesome.

PR #1459, opened 2025-09-05
Support for derived IInspectable COM interfaces in custom winmds

jevansaks:
We want to use CsWin32 to generate WinRT COM interfaces from header files that are not in the Windows SDK, but they derive from IInspectable. CsWin32 was not correctly handling these, especially in the mode of allowMarshalling=false. 

The problem was that we weren't being consistent about which MetadataReader we were using to resolve typedefs, methoddefs, and other things out of the winmd files when we crossed the winmd boundaries. In this PR we try to more clearly formalize how we access these things. Now we make sure to that the Generator we pass a TypeDefinition to (or whatever) has the owning MetadataReader for it. 

The biggest source of problems for this was HandleTypeHandleInfo, so I changed this to get its Generator when it's created. This meant I needed to pass the Generator down to the SignatureHandleProvider and make it bound to the Generator instead of a singleton. But now HandleTypeHandleInfo always has the Generator that it can use instead of needing the one from the TypeSyntaxSettings.

This PR adds a test that covers this scenario and we run through marshalling/non-marshalling and net472 and net8 combinations.


src/Microsoft.Windows.CsWin32/Generator.MetadataHelpers.cs
```diff
@@ -484,9 +502,13 @@ private bool IsManagedType(TypeDefinitionHandle typeDefinitionHandle)
 
                 throw new NotSupportedException();
             }
+            catch (GenerationFailedException)
+            {
+                throw;
+            }
```
AArnott, line 508: 🤔 Why avoid wrapping the exception with higher level info?

src/Microsoft.Windows.CsWin32/Generator.Com.cs
```diff
@@ -801,7 +803,7 @@ private unsafe (IReadOnlyList<MemberDeclarationSyntax> Members, IReadOnlyList<Ba
         if (populateVtblDeclared && this.IsFeatureAvailable(Feature.InterfaceStaticMembers) && !context.AllowMarshaling && GenerateCcwFor(originalIfaceName))
         {
             this.RequestComHelpers(context);
-            baseTypes.Add(SimpleBaseType(GenericName("IVTable").AddTypeArgumentListArguments(
+            baseTypes.Add(SimpleBaseType(GenericName($"{this.Win32NamespacePrefixString}.IVTable").AddTypeArgumentListArguments(
```
AArnott, line 806: Strictly speaking, I think this is an invalid `GenericName`. We should use a `ParseName` around this string before passing the result to `GenericName` so that we get a `QualifiedNameSyntax`.

src/Microsoft.Windows.CsWin32/MetadataCache.cs
```diff
@@ -34,7 +35,7 @@ internal MetadataFile GetMetadataFile(string path)
             }
 
             // New or updated file. Re-open.
-            this.metadataFiles.Add(path, metadataFile = new MetadataFile(path));
+            this.metadataFiles.Add(path, metadataFile = new MetadataFile(path, owner));
```
AArnott, line 38: ⚠️ This is only one return path, taken when there wasn't a cache hit. And in the process, we label it as belonging to _one_ particular `Generator` even though other generators that request the same path will get this same object without stamping it as belonging to them. `MetadataFile` instances are deliberately shared because they are expensive, yet readonly and thus shareable as long as the file hasn't changed. I think we need to find another way to pass an applicable `Generator` to those that need it.

src/Microsoft.Windows.CsWin32/MetadataFile.cs
```diff
@@ -34,6 +35,8 @@ internal MetadataFile(string path)
 
     internal DateTime LastWriteTimeUtc { get; }
 
+    internal Generator Generator { get; }
```
AArnott, line 38: ⚠️ As I point out in `MetadataCache`, we cannot add `Generator` specific fields to this class as it is shared across many generators concurrently.

src/Microsoft.Windows.CsWin32/MetadataIndex.cs
```diff
@@ -155,7 +155,7 @@ void PopulateNamespace(NamespaceDefinition ns, string? parentNamespace)
                                     fieldEnum.MoveNext();
                                     FieldDefinitionHandle fieldHandle = fieldEnum.Current;
                                     FieldDefinition fieldDef = mr.GetFieldDefinition(fieldHandle);
-                                    if (fieldDef.DecodeSignature(SignatureHandleProvider.Instance, null) is PrimitiveTypeHandleInfo { PrimitiveTypeCode: PrimitiveTypeCode.IntPtr or PrimitiveTypeCode.UIntPtr })
+                                    if (fieldDef.DecodeSignature(metadataFile.Generator.SignatureHandleProvider, null) is PrimitiveTypeHandleInfo { PrimitiveTypeCode: PrimitiveTypeCode.IntPtr or PrimitiveTypeCode.UIntPtr })
```
AArnott, line 158: As this appears to be the only reason we need `Generator` on the `MetadataFile`, hopefully this will be an easy coupling to undo.
Given we're only interested in _primitive_ type handles, which have no affinity to a particular winmd and therefore shouldn't need a particular `Generator`, and yet a full fledged `SignatureHandleProvider` *does* have a legit need to be affinitized to a particular `Generator`, I wonder if the solution is to create a special `SignatureHandleProvider`-like class that is just for indexes, and it will only support `PrimitiveTypeHandleInfo` since that's all we need. If the signature is _not_ such a handle, it can return null or whatever else is convenient. (We wouldn't want to _throw_ since this is a hot path.)

AArnott (changes requested):
A few touch-ups suggested, but the only real blocking issue is the mixing up of `MetadataFile` objects between generators, as my comments explain.

test/Microsoft.Windows.CsWin32.Tests/ExternalMetadata/NOTICE.md
```diff
@@ -3,3 +3,4 @@ This folder contains winmd files obtained from other sources which we use for te
 Metadata | Source
 --|--
 ServiceFabric.winmd | [youyuanwu/fabric-metadata](https://github.com/youyuanwu/fabric-metadata/raw/a1bcca6ad6f6a772c9e5ff4bdba80ae5e5f24cfc/.windows/winmd/ServiceFabric.winmd)
+CustomIInspectable.winmd | Generated using WinMDGenerator toolchain from [CustomIInspectable.idl](CustomIInspectable.idl)
```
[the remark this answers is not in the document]
jevansaks: I cleaned up the project and added it to a subdir so future maintainers can re-generate this if needed.

src/Microsoft.Windows.CsWin32/Generator.MetadataHelpers.cs
jevansaks: There are tests that are intentionally causing an exception and if we fall into the other catch block then we call ToTypeSyntax and immediately trigger the same exception and stack overflow. This enabled us to take the original GenerationFailedException and just bubble it out instead of unwrapping it and trying to re-generate the exact same message again (but failing).

jevansaks: If you can suggest a safer way to generate this exception message without potentially causing the exact same exception while doing so, I'm open to suggestions. :)

src/Microsoft.Windows.CsWin32/Generator.Com.cs
jevansaks: ParseName returns NameSyntax. I don't know how to turn that into a GenericName?

src/Microsoft.Windows.CsWin32/MetadataCache.cs
jevansaks: Fixed.

src/Microsoft.Windows.CsWin32/MetadataIndex.cs
jevansaks: Thanks. I added a PrimitiveSignatureHandleGenerator and that seems to have reduced the fallout from the change. Still running tests but looks good so far.

AArnott (changes requested):
Thanks. Much better. Almost there.
I'd fix it up myself but then I wouldn't be able to approve the PR 😒

src/Microsoft.Windows.CsWin32/Generator.Com.cs
AArnott: All I know for sure is it's weird. And if this works, that's fine.

test/CustomIInspectable/nuget.config
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
jevansaks: Yeah I did hit some restore problem. I'll delete this and hope it works later. :)

test/CustomIInspectable/CustomIInspectable.csproj
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.Build.NoTargets;Microsoft.Windows.WinmdGenerator/0.63.31-preview">
```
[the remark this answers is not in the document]
jevansaks: done

Issue #1472, opened 2025-10-04
Version 0.3.205 does not appear on GitHub


xtqqczze:
The NuGet package for version `0.3.205` of `Microsoft.Windows.CsWin32` is available on [NuGet.org](https://www.nuget.org/packages/Microsoft.Windows.CsWin32/0.3.205)), but this version does not appear on GitHub.

jevansaks:
Apologies! I've published the release notes now.

PR #1492, opened 2025-10-18
BuildTask mode should not generate types from InternalsVisibleTo referenced assemblies

jevansaks:
We were handling the case of not emitting types from referenced assemblies if they were _public_ but for InternalsVisibleTo we weren't. That's because the logic to check if we could see the type is based on the assemblyName in the CSharpCompilation we are building, and we passed in some placeholder thing.

The fix here is to plumb assemblyName through from the targets => task => generator => compilation. 

Also added a test to exercise this as well.


whiskhub:
Sorry to hijack this PR, but could you please ensure that for each new release on NuGet, a matching release is created [on GitHub](https://github.com/microsoft/CsWin32/releases)? See https://github.com/microsoft/CsWin32/issues/1472.

There are now multiple new releases on NuGet (in the last days), and we don't know what has changed. Does it already include the (eagerly awaited 🚀) AOT/ComWrappers support? Thanks for your efforts!

src/CsWin32Generator/Program.cs
```diff
@@ -106,6 +106,11 @@ public async Task<int> Main(
             AllowMultipleArgumentsPerToken = true,
         };
 
+        var assemblyNameOption = new Option<string?>("--assembly-name")
+        {
+            Description = "The name of the assembly being generated for.",
```
AArnott: This doc should specify that this is the _full_ name of the assembly, since the public key token is a critical part to the InternalsVisibleTo check.

jevansaks:
> Sorry to hijack this PR, but could you please ensure that for each new release on NuGet, a matching release is created [on GitHub](https://github.com/microsoft/CsWin32/releases)? See #1472.
> 
> There are now multiple new releases on NuGet (in the last days), and we don't know what has changed. Does it already include the (eagerly awaited 🚀) AOT/ComWrappers support? Thanks for your efforts!

I just published them. And yes, it does include that support. Please give it a try and let me know how it goes for you. :)

src/CsWin32Generator/Program.cs
jevansaks: I've added a new parameter for the key file.

integration-tests/buildtask-sdk/dependency/AssemblyInfo.cs
```diff
@@ -0,0 +1,2 @@
+﻿//[assembly: System.Runtime.CompilerServices.InternalsVisibleTo("buildtask-sdk")]
```
AArnott: Any reason to leave this first line, commented like this?

integration-tests/buildtask-sdk/buildtask-sdk.sln
```diff
@@ -0,0 +1,31 @@
+﻿
```
AArnott: Consider adding an slnx instead of an sln.

integration-tests/buildtask-sdk/dependency/AssemblyInfo.cs
jevansaks: No, will remove.

Issue #1496, opened 2025-10-20
PROPVARIANT (struct) usage fails with UseComSourceGenerators


BernhardMarconato:
## Actual behavior

COM interfaces that use `PROPVARIANT` or similar structs can not be used in a .NET JIT project, when CsWin32 is set to use `CsWin32RunAsBuildTask` and `UseComSourceGenerators`. A compiler error is raised:

`unsafe void GetProperty(winmdroot.Foundation.PROPERTYKEY* key, out winmdroot.System.Com.StructuredStorage.PROPVARIANT ppropvar);`
> SYSLIB1051	Runtime marshalling must be disabled in this project by applying the 'System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute' to the assembly to enable marshalling this type. The generated source will not handle marshalling of parameter 'ppropvar'. For types that are not supported by source-generated COM, the resulting function pointer will rely on the underlying runtime to marshal the specified type.

## Expected behavior

JIT (non-AOT) projects using CsWin32 with `CsWin32RunAsBuildTask` and `UseComSourceGenerators` can be used to generate COM interfaces using `PROPVARIANT` or similar structs without having to disable `DisableRuntimeMarshallingAttribute` for the entire assembly.

I understand this may be expected behavior right now, or due to limitations by .NET, but it would be nice if we didn't have to disable `DisableRuntimeMarshallingAttribute` for the entire assembly. Feel free to close the issue if this is not fixable.

## Repro steps

1. `NativeMethods.txt` content:
```
IShellItem2
```

2. `NativeMethods.json` content (if present):
```json
{
	"$schema": "https://aka.ms/CsWin32.schema.json",
	"comInterop": {"useComSourceGenerators": true}
}
```

3. Any of your own code that should be shared?
none required

### Context

- CsWin32 version: 0.3.217
- Target Framework: net8.0-windows10.0.26100.0
- CsWin32RunAsBuildTask = true

jevansaks:
@jkoritzinsky this warning is from the GeneratedComInterface source generator. What's your recommendation for devs who want to use this mode of cswin32?

jkoritzinsky:
I'd recommend one of the two following options:

1. If you're using the COM source generators, then you should be able to apply `[DisableRuntimeMarshalling]`. If you can't, then you might need to use that mode of CsWin32.
[...]

PR #1497, opened 2025-10-21
Update documentation for CsWin32RunAsBuildTask mode

### LINKED ISSUE #1493: Document new CsWin32RunAsBuildTask mode
The new CsWin32 mode to generate [GeneratedComInterface] and [LibraryImport] code is currently not well documented. I would have expected to find more information on https://microsoft.github.io/CsWin32/docs/getting-started.html (which is linked in the README).

It seems following is needed:
* Add `<CsWin32RunAsBuildTask>true</CsWin32RunAsBuildTask>` to csproj file
* Add `useComSourceGenerators ` in `comInterop` object in `NativeMethods.json` file

jevansaks:
Fixes #1493 #1496

### DIFF
--- a/docfx/docs/getting-started.md
+++ b/docfx/docs/getting-started.md
@@ -121,7 +121,7 @@ To keep CsWin32 generating the referred types you need, add them explicitly to `
 
 ### Support for trimming, AOT, and/or disabling the runtime marshaler
 
-**NEW!** CsWin32 now supports AOT by generating code from an MSBuild task rather than a source generator.
+**NEW!** CsWin32 now has improved support for NativeAOT by generating code from an MSBuild task rather than a source generator so that it can leverage other source generators that support `[LibraryImport]` and others.
 
 If you set one of these flags:
 


docfx/docs/getting-started.md
```diff
@@ -121,15 +121,36 @@ To keep CsWin32 generating the referred types you need, add them explicitly to `
 
 ### Support for trimming, AOT, and/or disabling the runtime marshaler
 
-Newer .NET runtime versions may fail for CsWin32 generated code when the application project builds with one or both of these properties set:
+**NEW!** CsWin32 now supports AOT by generating code from an MSBuild task rather than a source generator.
```
AArnott: nit: it supported AOT before. But the support is _better_ now. But this has historically been a sore point of contention so I'd rather not 'officially' document that there was not AOT support before. Can we reword this?

```suggestion
**NEW!** CsWin32 now has improved support for NativeAOT by generating code from an MSBuild task rather than a source generator so that it can leverage other source generators that support `[LibraryImport]` and others.
```

Issue #1500, opened 2025-10-21
New source generators mode does not generate working cocreateable classes (e.g. ShellLink)


Marv51:
## Actual behavior
The new source-generator based COM (`useComSourceGenerators: true`) generates a class with a (traditional) "ComImport" attribute for a cocreateable class like ShellLink. This class is incompatible with the interfaces generated with the same configuration.

## Expected behavior

The class and the interface are compatible with each other, both using the same technique.

## Repro steps

1. `NativeMethods.txt` content:
```txt
IPersistFile
IShellLinkW
ShellLink
```

2. `NativeMethods.json` content (if present):
```json
{
  "$schema": "https://aka.ms/CsWin32.schema.json",
  "comInterop": {
    "useComSourceGenerators": true
  }
}
```

3. Any of your own code that should be shared?
```csharp
var shellLink = new ShellLink();
var shellLinkW = (IShellLinkW)shellLink;

shellLinkW.SetPath("explorer.exe");
// some more calls
var persistFile = (IPersistFile)shellLink;
persistFile.Save(shortcut, true);
```

Both casts in that code snippet create this warning: SYSLIB1099 Casting between a `ComImport` type and a source-generated COM type is not supported

And the warning looks to be accurate the generated ShellLink:

```csharp
[Guid("00021401-0000-0000-C000-000000000046"),ComImport()]
[global::System.CodeDom.Compiler.GeneratedCode("Microsoft.Windows.CsWin32", "0.3.217+533aa1bddf.RR")]
internal partial class ShellLink
{
}
```

the generated IShellLinkW:
```csharp
[Guid("000214F9-0000-0000-C000-000000000046"),InterfaceType(ComInterfaceType.InterfaceIsIUnknown),GeneratedComInterface()]
[SupportedOSPlatform("windows5.1.2600")]
[global::System.CodeDom.Compiler.GeneratedCode("Microsoft.Windows.CsWin32", "0.3.217+533aa1bddf.RR")]
internal unsafe partial interface IShellLinkW
{
```

I believe that this line is responsible:
https://github.com/microsoft/CsWin32/blob/56ee24cb898c80c3b641837fdadc6ff6eb923802/src/Microsoft.Windows.CsWin32/Generator.Com.cs#L1318
But I do not know how to fix it.

For further context, I believe this issue created these classes: https://github.com/microsoft/CsWin32/issues/453

### Context

- CsWin32 version: 0.3.217
[...]

PR #1502, opened 2025-10-21
Handle CoCreateable classes in ComSourceGenerators mode

### LINKED ISSUE #1422: ShellLink doesn't work with nativeAOT
## Actual behavior

I want to publish my app with nativeAOT enabled.

In the app.csproj I set:
```xml
<PublishAot>true</PublishAot>
<BuiltInComInteropSupport>false</BuiltInComInteropSupport>
```

This is a Demo-WinUI Project that just contains a button that executes the code:

```csharp
var shellLink = new ShellLink();
var shellLinkW = (IShellLinkW.Interface)shellLink;
```

This throws an exception at runtime: Built-in COM has been disabled via a feature switch. See https://aka.ms/dotnet-illink/com for more information.

## Expected behavior

From reading the related issues in the repo, the code generated by CsWinRT should be safe for nativeAOT.

## Repro steps

```
IShellLinkW
ShellLink
```

```json
{
  "allowMarshaling": false
}
```

see above

### Context

- CsWin32 version: 0.3.183
- Target Framework: net8.0-windows10.0.19041.0

The generated code for ShellLink looks like this:
```csharp
		[Guid("00021401-0000-0000-C000-000000000046"),ComImport()]
		[global::System.CodeDom.Compiler.GeneratedCode("Microsoft.Windows.CsWin32", "0.3.183+73e6125f79.RR")]
		internal partial class ShellLink
		{
		}
```
The `ComImport` on a class looks problematic to me.

jevansaks:
COM source generators do not handle co-classes, you have to manually call CoCreateInstance. Per my proposal in #1500, I am making CsWin32 generate a class that looks like this:
```c#
[Guid("00021401-0000-0000-C000-000000000046")]
internal partial class ShellLink
{
    [Obsolete("Use CreateInstance<T> method instead.")]
    public ShellLink() { throw new NotSupportedException("Use CreateInstance<T> method instead."); }

    public static T CreateInstance<T>() where T : class
    {
        PInvoke.CoCreateInstance<T>(typeof(ShellLink).GUID, null, CLSCTX.CLSCTX_SERVER, out T ret).ThrowOnFailure();
        return ret;
    }
}
```

So the consumer would write code like this instead:
```c#
// BEFORE - ComInterop:
var shellLink = new ShellLink();
var shellLinkW = (IShellLinkW)shellLink;

// AFTER - ComSourceGenerators:
var shellLinkW = ShellLink.CreateInstance<IShellLinkW>();
```

One downside is this means it's hard to write consuming code that uses co classes and flips between the two modes. The problem is, in ComImport mode, you can't have _any_ methods on the class so we'd have to hide the factory method somewhere else which makes it less discoverable for the COM generators. In a future change we can use the "extension everything" .NET 10 feature to make this available in both modes.

I also generated a `CreateInstance<T>` for allowMarshaling=false mode because ... why not? The signature of the method returns HRESULT because it has to return a pointer out param. But this fixes #1422.

Fixes #1422, #1500.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.Com.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Com.cs
@@ -1319,7 +1319,7 @@ public partial class Generator
             .WithModifiers(classModifiers)
             .AddAttributeLists(AttributeList().AddAttributes(GUID(guid)).AddAttributes(canUseComImport ? [ComImportAttributeSyntax] : []));
 
-        if (!canUseComImport)
+        if (!canUseComImport && !this.Options.ComInterop.UseIntPtrForComOutPointers)
         {
             string obsoleteMessage = context.AllowMarshaling
                 ? $"COM source generators do not support direct instantiation of co-creatable classes. Use {name.Identifier}.CreateInstance<T> instead."
--- a/test/Microsoft.Windows.CsWin32.Tests/COMTests.cs
+++ b/test/Microsoft.Windows.CsWin32.Tests/COMTests.cs
@@ -500,7 +500,10 @@ public class COMTests : GeneratorTestBase
         // Check that it does not have the ComImport attribute.
         Assert.DoesNotContain(shellLinkType.AttributeLists, al => al.Attributes.Any(attr => attr.Name.ToString().Contains("ComImport")));
 
-        // Check that it contains a CreateInstance method
-        Assert.Contains(shellLinkType.DescendantNodes().OfType<MethodDeclarationSyntax>(), method => method.Identifier.Text == "CreateInstance");
+        if (!useIntPtrForComOutPtr)
+        {
+            // Check that it contains a CreateInstance method
+            Assert.Contains(shellLinkType.DescendantNodes().OfType<MethodDeclarationSyntax>(), method => method.Identifier.Text == "CreateInstance");
+        }
     }
 }


JeremyKuhne:
Have you considered just making the class a static class?

``` C#
[Guid("00021401-0000-0000-C000-000000000046")]
internal partial static class ShellLink
{
    public static T CreateInstance<T>() where T : class
    {
        PInvoke.CoCreateInstance<T>(typeof(ShellLink).GUID, null, CLSCTX.CLSCTX_SERVER, out T ret).ThrowOnFailure();
        return ret;
    }
}
```

Also, does it need to generic?

One other thing to consider is putting these statics on the actual interfaces when you're constrained to your `<DisableRuntimeMarshalling>` scenario:

``` C#
namespace Windows.Win32.UI.Shell;

public unsafe partial interface IShellLinkW
{
    public static Guid ClassGuid => new("00021401-0000-0000-C000-000000000046");

    public static IShellLinkW CreateInstance()
    {
        PInvoke.CoCreateInstance(ClassGuid, null, CLSCTX.CLSCTX_SERVER, out IShellLinkW ret).ThrowOnFailure();
        return ret;
    }
}
```

Having the helpers hanging off the interface might improve discovery. You could also include the static class as well to get people looking for it both ways?

jevansaks:
> Have you considered just making the class a static class?

I'm hoping to find a syntax that could work for both ComImport mode and not. ComImport has the constraint that class must be empty, so that does have me wondering if there should be a special static factory class somewhere else, where it could be static.

> One other thing to consider is putting these statics on the actual interfaces when you're constrained to your <DisableRuntimeMarshalling> scenario:

Unfortunately the declaration of coclass in win32metadata doesn't tell you anything other than the class name and GUID. What interfaces are associated with it is not described. So there's no way to know IShellLinkW would be the right interface.

Also with this approach you can ask for any interface on the thing, like: `ShellLink.CreateInstance<IPersistFile>()` would work too.

JeremyKuhne:
> Unfortunately the declaration of coclass in win32metadata doesn't tell you anything other than the class name and GUID. What interfaces are associated with it is not described. So there's no way to know IShellLinkW would be the right interface.

Bummer, might be a nice addition in the future.

jevansaks:
> Have you considered just making the class a static class?

Also, being a class with constructor that I can mark Obsolete will help folks find the new pattern when migrating from ComImport to GeneratedComInterface mode, so I think I'll stick with this pattern.

With .NET 10 I can do "extension everything" to make this a static extension method even on the ComImport type so folks can use same syntax before/after.

AaronRobinsonMSFT:
>         PInvoke.CoCreateInstance<T>(typeof(ShellLink).GUID, null, CLSCTX.CLSCTX_SERVER, out T ret).ThrowOnFailure();

This is a fine approach, but there is a slightly more AOT friendly manner.

``` csharp
[Guid("00021401-0000-0000-C000-000000000046")]
internal partial static class ShellLink
{
    private static ReadOnly<byte> CLSID_ShellLink = /* 00021401-0000-0000-C000-000000000046 as a byte array */;

    ...

    public static T CreateInstance<T>() where T : class
    {
        PInvoke.CoCreateInstance<T>(new Guid(CLSID_ShellLink), null, CLSCTX.CLSCTX_SERVER, out T ret).ThrowOnFailure();
        return ret;
    }
}
```

This is more pay-for-play. If a user does `typeof(ShellLink).GUID`, then metadata is kept via reflection, but if not then the additional metadata can be removed and the cost is only a simple read only byte array.

jevansaks:
> This is more pay-for-play.

If no one calls CreateInstance, won't the trimmer see it's uncalled and also remove the reference to typeof(ShellLink).GUID and thus realize it doesn't need to keep metadata? Or is the determination of what metadata to keep made before trimming?

jevansaks:
But I'm ok with adding a static. Is there a reason the static should be ReadOnlySpan<byte> and then construct the GUID in the function versus the static is:
```
private static Guid CLSID_ShellLink => new Guid(0x00021401, 0x0000, 0x0000, 0xC0, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x46);
```
?

PR #1504, opened 2025-10-22
Fix analyzer test break in devdiv AzDO account

AArnott:
Fix analyzer tests to pull packages from our nuget.config feed

This fixes a Azure Pipeline break in the devdiv project where network isolation is enabled.


jevansaks (approved):
Looks good. I also queued the "unofficial" pipeline to make sure that works too. We can merge after that passes.

jevansaks:
Pipeline failed because the package can't be pulled from upstream. But I think if I manually save to feed it will work. 🤞

jevansaks:
I pushed some changes to the branch to qualify other ReferenceAssemblies usage with the "WithNugetConfigFilePath" qualifier.

manodasanW (approved):

PR #1511, opened 2025-10-28
Improve projection of MemorySize-d and optional ref/out parameters

### LINKED ISSUE #1446: Friendly overload doesn't turn pointer into managed ref
## Expected behavior
The friendly overload signature doesn't contain any raw pointers

## Actual behavior
Friendly overload signature is `uint MsiRecordGetString(SafeHandle hRecord, uint iField, Span<char>szValueBuf, uint* pcchValueBuf)`. Notice that the last parameter is a pointer. The [documentation page](https://learn.microsoft.com/en-us/windows/win32/api/msiquery/nf-msiquery-msirecordgetstringw) says that the last parameter is both in and out pointer, so it should become `ref` in the friendly signature

## Repro steps

`NativeMethods.txt` content:
```
MsiRecordGetString
```

### Context

- CsWin32 version: `0.3.183`
- Target Framework: `net10.0`

### LINKED ISSUE #1453: How to make the generated interface parameters become Intptr? I don't want to use unsafe code to pass parameters
Image

### LINKED ISSUE #1456: The SafeHandle overload for `InitializeAcl()` is incorrect.
## Actual behavior

This P/Invoke wrapper is sending a pointer to a fixed address of the out'd ACL variable when this P/Invoke expects a buffer allocated with LocalAlloc be allocated and sent through first.
```c#
internal static unsafe winmdroot.Foundation.BOOL InitializeAcl(out winmdroot.Security.ACL pAcl, uint nAclLength, winmdroot.Security.ACE_REVISION dwAclRevision)
{
	fixed (winmdroot.Security.ACL* pAclLocal = &pAcl)
	{
		winmdroot.Foundation.BOOL __result = PInvoke.InitializeAcl(pAclLocal, nAclLength, dwAclRevision);
		return __result;
	}
}
```

## Expected behavior

This wrapper should be doing a LocalAlloc allocation to the specified size of `nAclLength`, sending a pointer for this through to `InitializeAcl()`, then the `pAcl` parameter should be out'ing a `LocalFreeSafeHandle` type to work with.

## Repro steps

```
InitializeAcl
```

N/A

### Context

- CsWin32 version: Latest
- Win32Metadata version: N/A
- Target Framework: net472
- `LangVersion`: N/A

jevansaks:
This change addresses a few problems with projection of certain parameters:

### 1. Improve projection of `[optional, out]` parameters

Currently cswin32 projects `[optional, out]` parameters as pointer, which makes it harder to work with. After discussion in #1446, I propose the following new behavior:

Some parameters in win32 are `[optional, out]` or `[optional, in, out]`. C# does not have an idiomatic way to represent this concept, so for any method that
has such parameters, CsWin32 will generate two versions: one with all `ref` or `out` parameters included, and one with all such parameters omitted. For example:

```c#
// Omitting the optional parameter:
IsTextUnicode(buffer);

// Passing ref for optional parameter:
IS_TEXT_UNICODE_RESULT result = default;
IsTextUnicode(buffer, ref result);
```

### 2. Project struct-typed `[MemorySize]` parameters as `Span<byte>`

The handling of struct-typed `[MemorySize]` parameters in cswin32 is in some cases incorrect: #1456, and there's no way to correctly call the friendly method since you cannot pass a buffer larger than the struct. In such cases, these parameters are like FlexibleArrays but the struct doesn't have an embedded variable-sized array, the struct is more like the header of some variable-sized payload. A good example is InitializeAcl (from the linked issue). 

The ideal would be `Span<ACL>` where you can express the size of the Span in a number of bytes. However, Span has no way to do that, Span is always a whole number of elements. So, we need to type these parameters as `Span<byte>`. I added the following blurb to the getting-started docs to explain how these work:

In the Win32 APIs there are many functions where one parameter is a buffer (`void*` or `byte*`) and another parameter is the size of that buffer. When generating for a target framework that supports Spans, there will be overloads of these functions that take a `Span<byte>` which represents both of these parameters, since a Span refers to a chunk of memory and a length. For example, an API like [IsTextUnicode](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-istextunicode) has a `void*` parameter whose length is described by the iSize parameter in the native signature. The CsWin32 projection of this method will be:

```c#
BOOL IsTextUnicode(ReadOnlySpan<byte> lpv, ref IS_TEXT_UNICODE_RESULT lpiResult)
```

Instead of passing the buffer and length separately, in this projection you pass just one parameter. Span is a flexible type with many things that can be converted to it safely. You will also see Span parameters for things that may look like a struct but are variable sized. For example, [InitializeAcl](https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-initializeacl) looks like it returns an ACL struct but the parameter is annotated with a `[MemorySize]` attribute in the metadata, indicating it is variable-sized based on another parameter. Thus, the cswin32 projection of this method will project this parameter as a `Span<byte>` since the size of the parameter is variable:

```c#
// The cswin32 signature:
static BOOL InitializeAcl(Span<byte> pAcl, ACE_REVISION dwAclRevision) { ... }
```

And you would call this by creating a buffer to receive the ACL. Then, after the call you can reinterpret the buffer as an ACL:
```c#
// Make a buffer
Span<byte> buffer = new byte[CalculateAclSize(...)];
InitializeAcl(buffer, ACE_REVISION.ACL_REVISION);

// The beginning of the buffer is an ACL, so cast it to a ref:
ref ACL acl = ref MemoryMarshal.AsRef<ACL>(buffer);

// Or treat it as a Span:
Span<ACL> aclSpan = MemoryMarshal.Cast<byte, ACL>(buffer);
```

CsWin32 will also generate a struct-typed parameter for convenience but this overload will pass `sizeof(T)` for the length parameter to the underlying Win32 API, so this only makes sense in some overloads such as [SHGetFileInfo](https://learn.microsoft.com/windows/win32/api/shellapi/nf-shellapi-shgetfileinfow) where the parameter has an annotation indicating it's variable-sized, but the size is only ever `sizeof(SHFILEINFOW)`:

```c#
// Span<byte> overload:
static nuint SHGetFileInfo(string pszPath, FILE_FLAGS_AND_ATTRIBUTES dwFileAttributes, Span<byte> psfi, SHGFI_FLAGS uFlags)
// ref SHGETFILEINFOW overload:
static nuint SHGetFileInfo(string pszPath, FILE_FLAGS_AND_ATTRIBUTES dwFileAttributes, ref SHFILEINFOW psfi, SHGFI_FLAGS uFlags)
```

Fixes #1446, Fixes #1456, Fixes #1453.

### DIFF
--- a/test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
+++ b/test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
@@ -57,6 +57,7 @@ public partial class CsWin32GeneratorTestsBase : GeneratorTestBase
 
     protected async Task InvokeGeneratorAndCompile(string testCase, TestOptions options = TestOptions.None)
     {
+        this.compilation = this.compilation.WithAssemblyName(this.assemblyName);
         this.compilation = this.compilation.AddReferences(this.additionalReferences.Select(x => MetadataReference.CreateFromFile(x)));
 
         string outputPath = this.GetTestCaseOutputDirectory(testCase);
--- a/src/Microsoft.Windows.CsWin32/Generator.Handle.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Handle.cs
@@ -70,7 +70,9 @@ public partial class Generator
                 return safeHandleType;
             }
 
-            if (this.IsTypeAlreadyFullyDeclared($"{this.Namespace}.{safeHandleType}"))
+            // Bail out early if someone already made the SafeHandle type
+            string safeHandleFullyQualifiedName = $"{this.Namespace}.{safeHandleClassName}";
+            if (this.IsTypeAlreadyFullyDeclared(safeHandleFullyQualifiedName))
             {
                 return safeHandleType;
             }
--- a/src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.TypeDef.cs
@@ -278,7 +278,7 @@ public partial class Generator
             yield return ConstructorDeclaration(structName.Identifier)
                 .AddModifiers(TokenWithSpace(this.Visibility))
                 .AddParameterListParameters(Parameter(valueParameter.Identifier).WithType(IntPtrTypeSyntax.WithTrailingTrivia(TriviaList(Space))))
-                .WithInitializer(ConstructorInitializer(SyntaxKind.ThisConstructorInitializer).AddArgumentListArguments(Argument(CastExpression(fieldType, valueParameter))))
+                .WithInitializer(ConstructorInitializer(SyntaxKind.ThisConstructorInitializer).AddArgumentListArguments(Argument(UncheckedExpression(CastExpression(fieldType, valueParameter)))))
                 .WithBody(Block());
         }
 
--- a/src/Microsoft.Windows.CsWin32/templates/HRESULT.cs
+++ b/src/Microsoft.Windows.CsWin32/templates/HRESULT.cs
@@ -38,7 +38,7 @@
 partial struct HRESULT
 {
 	public static implicit operator uint(HRESULT value) => (uint)value.Value;
-	public static explicit operator HRESULT(uint value) => new HRESULT((int)value);
+	public static explicit operator HRESULT(uint value) => new HRESULT(unchecked((int)value));
 
 	[DebuggerBrowsable(DebuggerBrowsableState.Never)]
 	internal bool Succeeded => this.Value >= 0;
--- a/src/Microsoft.Windows.CsWin32/templates/NTSTATUS.cs
+++ b/src/Microsoft.Windows.CsWin32/templates/NTSTATUS.cs
@@ -26,9 +26,9 @@
 partial struct NTSTATUS
 {
 	public static implicit operator uint(NTSTATUS value) => (uint)value.Value;
-	public static explicit operator NTSTATUS(uint value) => new NTSTATUS((int)value);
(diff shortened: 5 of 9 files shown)


src/Microsoft.Windows.CsWin32/templates/CharSet_Workaround.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Windows.Win32.System.Runtime.InteropServices
+﻿namespace System.Runtime.InteropServices
```
manodasanW: Does this not conflict with https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.charset?view=net-9.0

test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
```diff
@@ -161,14 +171,23 @@ protected async Task CompileGeneratedFilesWithSourceGenerators(string outputPath
             syntaxTrees.Add(syntaxTree);
         }
 
-        // Add the assembly attribute for DisableRuntimeMarshalling
-        string disableRuntimeMarshallingSource = @"
+        bool canDisableRuntimeMarshalling = this.tfm is null ||
+            this.tfm.StartsWith("net6.0", StringComparison.OrdinalIgnoreCase) ||
```
manodasanW: docs seem to indicate the attribute was added in .NET 7: https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.disableruntimemarshallingattribute?view=net-9.0

PR #1512, opened 2025-10-28
Simplify decimal conversions

### LINKED ISSUE #1455: Implement decimal to DECIMAL conversion on non-.NET platforms
The .NET `decimal` type doesn't expose a proper API for extracting raw values without allocating an `int[]` unless you target .NET.
But .NET Standard targets can do this too by leveraging the field layout of the struct being intentionally the same between the two.

```cs
namespace Windows.Win32.Foundation
{
	/// <content>
	/// Adds implicit conversion from decimal to DECIMAL.
	/// </content>
	internal partial struct DECIMAL
	{
		public static unsafe implicit operator DECIMAL(decimal value)
		{
			// DECIMAL is layout-compatible with decimal.
			return *(DECIMAL*)&value;
		}
	}
}
```

Also, the converter that _is_ emitted is buggy:

```cs
public static implicit operator decimal(DECIMAL value)
{
	return new decimal(
(int)value.Anonymous2.Anonymous.Lo32, (int)value.Anonymous2.Anonymous.Mid32, (int)value.Hi32, value.Anonymous1.Anonymous.sign == 0x80, value.Anonymous1.Anonymous.scale);
}
```

This throws `OverflowException` as it casts `uint` fields to `int` values without `unchecked`.

jevansaks:
Per suggestion in #1455. Tests already existed to exercise decimal conversions and they still pass.

Fixes #1455.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/templates/DECIMAL.cs
+++ b/src/Microsoft.Windows.CsWin32/templates/DECIMAL.cs
@@ -12,7 +12,5 @@
 		return *(decimal*)&value;
 	}
 
-#if NET5_0_OR_GREATER
 	public static implicit operator DECIMAL(decimal value) => new DECIMAL(value);
-#endif
 }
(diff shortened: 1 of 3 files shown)


jevansaks:
There is such a conversion operator but it was `#if NET5_0_OR_GREATER` when I got here. I took the `#if` away and the project compiles and the NET 3.5 tests pass. Do you have any history on why that conversion operator was excluded from .NET Framework previously?

Issue #1229, opened 2024-07-01
MFTEnumEx signature incorrect


jimm98y:
## Actual behavior

The signature of the generated MFTEnumEx is not correct, causing a crash. This used to be working in 0.3.49-beta, it is broken since the latest version 0.3.106. The new generated function in the latest version is:
```cs
public static unsafe winmdroot.Foundation.HRESULT MFTEnumEx(global::System.Guid guidCategory, winmdroot.Media.MediaFoundation.MFT_ENUM_FLAG Flags, winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO? pInputType, winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO? pOutputType, out winmdroot.Media.MediaFoundation.IMFActivate pppMFTActivate, out uint pnumMFTActivate)
```

In 0.3.49-beta which was working it is:
```cs
public static unsafe winmdroot.Foundation.HRESULT MFTEnumEx(global::System.Guid guidCategory, winmdroot.Media.MediaFoundation.MFT_ENUM_FLAG Flags, winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO? pInputType, winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO? pOutputType, out winmdroot.Media.MediaFoundation.IMFActivate[] pppMFTActivate, out uint pnumMFTActivate)
```

You can notice the IMFActivate[] parameter is now typed as IMFActivate which is why it's not working. The definition looks correct:

```c
STDAPI
MFTEnumEx(
    _In_                                GUID                            guidCategory,
    _In_                                UINT32                          Flags,
    _In_opt_                            const MFT_REGISTER_TYPE_INFO*   pInputType,
    _In_opt_                            const MFT_REGISTER_TYPE_INFO*   pOutputType,
    _Outptr_result_buffer_(*pnumMFTActivate) IMFActivate***                 pppMFTActivate,
    _Out_                               UINT32*                         pnumMFTActivate
);
```

## Expected behavior

The function definition should include IMFActivate[] and calling it should not crash the app.

## Repro steps

1. `NativeMethods.txt` content:
```
mf.*
mfplat.*
MF*
_MFT_INPUT_STATUS_FLAGS
_MFT_PROCESS_OUTPUT_FLAGS
_MFT_PROCESS_OUTPUT_STATUS
[...]

PR #1515, opened 2025-10-29
Fixed incorrect parameters generated for pointer-to-pointer types #1229

jimm98y:
After this change, it looks like it generates the correct array type for the MFTEnumEx:

```cs
[LibraryImport("MFPlat.dll")]
internal static unsafe partial winmdroot.Foundation.HRESULT MFTEnumEx(global::System.Guid guidCategory, [global::System.Runtime.InteropServices.Marshalling.MarshalUsing(typeof(global::Windows.Win32.Media.MediaFoundation.InteropServices.MFT_ENUM_FLAGToU4Marshaler))] winmdroot.Media.MediaFoundation.MFT_ENUM_FLAG Flags, [Optional] winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO* pInputType, [Optional] winmdroot.Media.MediaFoundation.MFT_REGISTER_TYPE_INFO* pOutputType, out winmdroot.Media.MediaFoundation.IMFActivate[] pppMFTActivate, uint* pnumMFTActivate);

```

### DIFF
--- a/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs
+++ b/src/Microsoft.Windows.CsWin32/PointerTypeHandleInfo.cs
@@ -42,7 +42,7 @@ internal record PointerTypeHandleInfo(TypeHandleInfo ElementType) : TypeHandleIn
                 parameterAttributes).Type));
         }
 
-        TypeSyntaxAndMarshaling elementTypeDetails = this.ElementType.ToTypeSyntax(inputs with { PreferInOutRef = false }, forElement, customAttributes, parameterAttributes);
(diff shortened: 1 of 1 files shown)


jimm98y:
Unfortunately, this is not going to work:

error SYSLIB1051: Runtime marshalling must be disabled in this project by applying the 'System.Runtime.CompilerServices.DisableRuntimeMarshallingAttribute' to the assembly to enable marshalling this type. The generated source will not handle marshalling of parameter 'Value'.
error SYSLIB1051: Marshalling an array from unmanaged to managed requires either the 'SizeParamIndex' or 'SizeConst' fields to be set on a 'MarshalAsAttribute' or the 'ConstantElementCount' or 'CountElementName' properties to be set on a 'MarshalUsingAttribute'. The generated source will not handle marshalling of parameter 'pppMFTActivate'.

PR #1545, opened 2025-11-13
Generate struct wrapper around function pointer to make a native delegate typedef

### LINKED ISSUE #778: Generate wrapper structs for unmanaged delegates
APIs take native function pointers, which are defined in the metadata, but we have no wrapping projection for them.

For example, in the Win32Metadata:

``` C#
[StructLayout(LayoutKind.Auto, CharSet = CharSet.Auto)]
[UnmanagedFunctionPointer(CallingConvention.Winapi)]
public delegate LRESULT WNDPROC([In] HWND param0, [In] uint param1, [In] WPARAM param2, [In] LPARAM param3);
```

Create a "handle" struct like we do for `HANDLE` objects.  Here is how I've wrapped this manually:

``` C#
public unsafe readonly struct WNDPROC
{
    public readonly delegate* unmanaged<HWND, uint, WPARAM, LPARAM, LRESULT> Value;

    public WNDPROC(delegate* unmanaged<HWND, uint, WPARAM, LPARAM, LRESULT> value) => Value = value;
    public bool IsNull => Value is null;

    public static implicit operator WNDPROC(delegate* unmanaged<HWND, uint, WPARAM, LPARAM, LRESULT> value)
        => new(value);
    public static implicit operator delegate* unmanaged<HWND, uint, WPARAM, LPARAM, LRESULT>(WNDPROC value)
        => value.Value;
}
```

With that you should be able to use the struct in the APIs.

### LINKED ISSUE #1542: CsWin32RunAsBuildTask fails to generate delegate structs (e.g. HOOKPROC, WNDPROC, SUBCLASSPROC)
## Actual behavior

When `CsWin32RunAsBuildTask` is set to `true`, structs declared in `NativeMethods.txt` are not generated. Errors appear during build: 
> *The type or namespace name 'HOOKPROC' does not exist in the namespace 'Windows.Win32.UI.WindowsAndMessaging' (are you missing an assembly reference?)*  

Tested with `HOOKPROC`, `WNDPROC`, and `SUBCLASSPROC` — none of them are generated.

If `CsWin32RunAsBuildTask` is set to `false`, everything works as expected and the structs are generated correctly.

## Expected behavior

Structs should be generated in a NativeAOT app when `CsWin32RunAsBuildTask` is set to `true`.

## Repro steps

```
HOOKPROC
WNDPROC
SUBCLASSPROC
```

```json
{
  "allowMarshaling": false,
  "public": true
}
```

csproj content:
```
	<PublishAot>true</PublishAot>
	<PublishTrimmed>true</PublishTrimmed>
```

### Context

- CsWin32 version: 0.3.242
- Target Framework:net8.0-windows10.0.26100.0
- `LangVersion` (if explicitly set by project): preview

jevansaks:
Per @JeremyKuhne's suggestion in #778, generate a struct wrapper to represent typedefs like WNDPROC and TIMERPROC. You can now also request these types and get a reasonable definition out.

The one problem is that function pointers do not seem to work via implicit conversions so this is slightly worse for function pointers. I am going to leave this as a draft until we can figure that out.

Fixes #1542, Fixes #778.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.Delegate.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Delegate.cs
@@ -176,7 +176,7 @@ public partial class Generator
             .WithExpressionBody(ArrowExpressionClause(AssignmentExpression(SyntaxKind.SimpleAssignmentExpression, IdentifierName("Value"), IdentifierName("value"))))
             .WithSemicolonToken(SemicolonWithLineFeed);
 
-        // public <Name> Null => default;
+        // public static <Name> Null => default;
         PropertyDeclarationSyntax nullProperty = PropertyDeclaration(IdentifierName(name).WithTrailingTrivia(Space), Identifier("Null"))
             .AddModifiers(TokenWithSpace(SyntaxKind.PublicKeyword), TokenWithSpace(SyntaxKind.StaticKeyword))
             .WithExpressionBody(ArrowExpressionClause(LiteralExpression(SyntaxKind.DefaultLiteralExpression)))


jevansaks:
I've updated this PR so that it generates the wrapper struct but it still leaves the parameters and struct fields as function pointers. If you explicitly request "WNDPROC" or "TIMERPROC" then it will generate the struct, and you can use it to wrap your function pointers. If we ever get improvements in C# parsing, we can revisit (FWIW I filed https://github.com/dotnet/roslyn/issues/81234 and it was summarily closed).

manodasanW (approved):

PR #1565, opened 2025-11-24
Switch CsWin32RunAsBuildTask to EmitSingleFile by default (for VS incremental scenario)

jevansaks:
@drewnoakes recommended against the approach in #1557, and we can do about as well by just changing the tool to generate a single file by default.

This also can help with discoverability because F12 will take you into the single generated file and let you find all the other types that are generated more easily.

Drew also suggested that a glob would work to avoid the DTB approach but that complicates the targets a bit so I'm going to avoid doing that for now.

### DIFF
--- a/src/CsWin32Generator/Program.cs
+++ b/src/CsWin32Generator/Program.cs
@@ -264,6 +264,12 @@ public partial class Program
             options.ComInterop.UseComSourceGenerators = true;
         }
 
+        // Set EmitSingleFile to true by default if unspecified.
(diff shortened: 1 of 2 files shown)


manodasanW (approved):

jevansaks:
@manodasanw I changed this to make the EmitSingleFile field in the GeneratorOptions nullable so I don't have to parse the json twice to see if it's unset. I resisted doing that initially but then I found I wasn't passing all the same parsing options so this is more robust anyway.

drewnoakes (approved):

PR #1589, opened 2025-12-05
Add [Optional] on optional params

### LINKED ISSUE #652: FwpmFilterAdd0: optional parameter cannot be null
## Actual behavior

The last two optional parameters of FwpmFilterAdd0 cannot be null

## Expected behavior

The last two optional parameters of FwpmFilterAdd0 can be null

## Repro steps

```
FwpmFilterAdd0
OpenProcess
QueryFullProcessImageName
```

### Context

- CsWin32 version: 0.2.10-beta
- Target Framework: .NET 6

### LINKED ISSUE #1580: `FwpmProviderAdd0` last parameter is not treated as optional
## Actual behavior
Neither in friendly nor in native overload there is a clue that parameter is optional

## Expected behavior
In native overload the last patameter is marked as `[Optional]`, in friendly overload it has a default value (`winmdroot.Security.PSECURITY_DESCRIPTOR sd = default`)

## Repro steps

```
FwpmProviderAdd0
```

### Context

- CsWin32 version: `0.3.257`
- Win32Metadata version (if explicitly set by project): `68.0.4-preview`

It seems that in metadata this parameter is annotated correctly: `[Optional][In] PSECURITY_DESCRIPTOR sd`

jevansaks:
There is a C# feature that if you have `[Optional]` on parameters that callers can omit those parameters and the compiler will automatically pass `default` for those parameters. This is _almost_ like providing `= default` values on the parameters except that it _also_ works for any parameter, not just trailing parameters. 

What this means is that we can put `[Optional]` on any parameter that is optional which doesn't also have an `out` or `ref` modifier on it, and that will give callers a little more flexibility and communicate which parameters are optional / can be omitted.

Fixes #1580 , Fixes #652

### DIFF
--- a/test/CsWin32Generator.Tests/CsWin32GeneratorTests.cs
+++ b/test/CsWin32Generator.Tests/CsWin32GeneratorTests.cs
@@ -219,7 +219,8 @@ public partial class CsWin32GeneratorTests : CsWin32GeneratorTestsBase
         ["ITypeInfo", "GetNames", "this winmdroot.System.Com.ITypeInfo @this, int memid, Span<winmdroot.Foundation.BSTR> rgBstrNames, out uint pcNames"],
         ["EnumProcessModules", "EnumProcessModules", "SafeHandle hProcess, Span<byte> lphModule, out uint lpcbNeeded"],
         ["Windows.Win32.NetworkManagement.WindowsFilteringPlatform.FwpmProviderAdd0", "FwpmProviderAdd0", "SafeHandle engineHandle, in winmdroot.NetworkManagement.WindowsFilteringPlatform.FWPM_PROVIDER0 provider, [Optional] winmdroot.Security.PSECURITY_DESCRIPTOR sd"],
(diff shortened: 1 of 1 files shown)


DoctorKrolic:
Can you please also add test for `FwpmEngineOpen0`? As per https://github.com/microsoft/win32metadata/issues/2153 the first parameter in native overload didn't have `[Optional]` on it. It is not tagged `[Reserved]`, but the native overload still lacks `[Optional]` attribute. This is either already fixed with your changes or needs some additional tweaks. Either way, it is a good test case verifying that a combination of `[Reserved]` and `[Optional]` still puts `[Optional]` in the native overload

jevansaks:
I added a test for that function. In the latest win32metadata, that parameter is [Reserved] so it's always omitted on the friendly overload.

DoctorKrolic:
I meant a test for the native overload. It is a good thing that this parameter is omitted from the friendly one, but the native should still have this parameter as `[Optional]`, so it can be skipped when using named argiments

jevansaks:
Oh, sure. Updated that test to check the native signature instead.

Issue #1078, opened 2023-10-24
Support [RAIIFree] on output parameters


AArnott:
## Actual behavior

Given the following metadata definition of `WlanOpenHandle`:

```cs
public unsafe static extern uint WlanOpenHandle([In] uint dwClientVersion, [Optional][Reserved] void* pReserved, [Out] uint* pdwNegotiatedVersion, [Out][RAIIFree("WlanCloseHandle")] HANDLE* phClientHandle);
```

CsWin32 generates the following friendly overload:

```cs
internal static unsafe uint WlanOpenHandle(uint dwClientVersion, out uint pdwNegotiatedVersion, winmdroot.Foundation.HANDLE* phClientHandle)
```

## Expected behavior

I expect the last parameter to be an `out SafeHandle`-derived type that calls `WlanCloseHandle` when released.

PR #1603, opened 2026-01-01
Generate `SafeHandle` when freeing method accepts additional reserved parameters

DoctorKrolic:
Fixes: https://github.com/microsoft/CsWin32/issues/1078

We already handle out handle parameters annotated with `[RAIIFree]`. However, in this particular case `WlanCloseHandle` function accepts 2 parameters, which usually blocks us from turning raw handle into a `SafeHandle`. But the second parameter of `WlanCloseHandle` is a `[Reserved]` one, so we don't actually need to pass a value to it. So the fix is to recognize such cases and handle still allow `SafeHandle` to be used

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.Handle.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.Handle.cs
@@ -185,7 +185,11 @@ public partial class Generator
                     IdentifierName(renamedReleaseMethod ?? releaseMethod)),
                 ArgumentList().AddArguments(releaseHandleArgument));
             BlockSyntax? releaseBlock = null;
-            bool releaseMethodIsUnsafe = false;
+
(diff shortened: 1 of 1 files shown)


DoctorKrolic:
@jevansaks PTAL

jevansaks:
@doctorkrolic Does the whole solution build for you? When I checked this out and built locally, I got build breaks that SysFreeStringSafeHandle is no longer generated. Seems like this might have broken generation of some other SafeHandles.

DoctorKrolic:
It seems that CI failure is not real

jevansaks:
I'll try to re-run. We may need to scale back on some of the tests we run in the lab because of memory constraints. We're always just on the edge and some changes seem to cause the agents to blow up.

@AArnott any thoughts on how to request different agents with more memory for running the tests?

Issue #1430, opened 2025-06-17
Resource leak if allocation of `SafeHandle` fails


carlreinke:
## Actual behavior

The generated code leaks the resource if allocation of the `SafeHandle` instance causes OOM.  E.g.
```C#
internal static unsafe winmdroot.Devices.DeviceAndDriverInstallation.CONFIGRET CM_Open_DevNode_Key(uint dnDevNode, uint samDesired, uint ulHardwareProfile, uint Disposition, out Microsoft.Win32.SafeHandles.SafeRegistryHandle phkDevice, uint ulFlags)
{
    winmdroot.System.Registry.HKEY phkDeviceLocal;
    winmdroot.Devices.DeviceAndDriverInstallation.CONFIGRET __result = PInvoke.CM_Open_DevNode_Key(dnDevNode, samDesired, ulHardwareProfile, Disposition, &phkDeviceLocal, ulFlags);
    phkDevice = new Microsoft.Win32.SafeHandles.SafeRegistryHandle(phkDeviceLocal, ownsHandle: true);  // OOM
    return __result;
}
```

## Expected behavior

The generated code should allocate the `SafeHandle` instance before the P/Invoke and use [Marshal.InitHandle](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.marshal.inithandle) to initialize the handle after the P/Invoke.  E.g.
```C#
internal static unsafe winmdroot.Devices.DeviceAndDriverInstallation.CONFIGRET CM_Open_DevNode_Key(uint dnDevNode, uint samDesired, uint ulHardwareProfile, uint Disposition, out Microsoft.Win32.SafeHandles.SafeRegistryHandle phkDevice, uint ulFlags)
{
    phkDevice = new Microsoft.Win32.SafeHandles.SafeRegistryHandle();
    winmdroot.System.Registry.HKEY phkDeviceLocal;
    winmdroot.Devices.DeviceAndDriverInstallation.CONFIGRET __result = PInvoke.CM_Open_DevNode_Key(dnDevNode, samDesired, ulHardwareProfile, Disposition, &phkDeviceLocal, ulFlags);
    System.Runtime.InteropServices.Marshal.InitHandle(phkDevice, phkDeviceLocal);
    return __result;
}
```

See also the [API review for Marshal.InitHandle](https://github.com/dotnet/runtime/issues/46830#issuecomment-761148142).

### Context

- CsWin32 version: 0.3.183
- Target Framework: `net8.0`

[...]

PR #1613, opened 2026-01-13
Use `Marshal.InitHandle` API to avoid memory leak when OOM happens

DoctorKrolic:
Fixes: https://github.com/microsoft/CsWin32/issues/1430

### DIFF
--- a/src/Microsoft.Windows.CsWin32/FastSyntaxFactory.cs
+++ b/src/Microsoft.Windows.CsWin32/FastSyntaxFactory.cs
@@ -194,7 +194,7 @@ internal static class FastSyntaxFactory
 
     internal static InitializerExpressionSyntax InitializerExpression(SyntaxKind kind, SeparatedSyntaxList<ExpressionSyntax> expressions) => SyntaxFactory.InitializerExpression(kind, OpenBrace, expressions, CloseBrace);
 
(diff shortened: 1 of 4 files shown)


jevansaks:
I checked out this branch locally and started running tests and I see a few test failures -- "GeneratorTests.InterestingAPIs":

> Windows.Win32.NativeMethods.g.cs(64,33): error CS1503: Argument 2: cannot convert from 'Windows.Win32.Security.Cryptography.NCRYPT_KEY_HANDLE' to 'nint'

@doctorkrolic If you can, run tests locally in Test Explorer. _Ideally_ the full set, but the "Full" generation tests can take a long time so I'm ok if you just run the "No Traits" set before putting up a PR:

image

Click the "Group By" button in task explorer to group by Traits and then run the "No Traits" set:

image

DoctorKrolic:
> run tests locally in Test Explorer

Is there a apecific reason why CI is blocked here from "untrusted" contributors? I am used to rely on CI to uncover test failures and all repos in dotnet org I worked with have their CI open to any contributor. It seems very inconvinient, that you have to manually trigger CI run every time I make a commit

jevansaks:
> Is there a apecific reason why CI is blocked here from "untrusted" contributors? I am used to rely on CI to uncover test failures and all repos in dotnet org I worked with have their CI open to any contributor. It seems very inconvinient, that you have to manually trigger CI run every time I make a commit

I defer to @aarnott who set up the pipelines. This may be because the pipelines are in ADO and if they were GitHub actions it would be safer to run them automatically? Not sure. It's like this in a lot of repos I work in.

jevansaks (approved):

PR #1624, opened 2026-01-22
Use concrete `SafeHandle` type when accepting a handle in friendly signature other than `HANDLE`

DoctorKrolic:
Makes friendly signatures more strongly typed

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
@@ -483,9 +483,13 @@ public partial class Generator
                         ExpressionStatement(InvocationExpression(MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression, origName, IdentifierName(nameof(SafeHandle.DangerousRelease))))))
                     .WithCloseParenToken(TokenWithLineFeed(SyntaxKind.CloseParenToken)));
 
-                // Accept the SafeHandle instead.
-                parameters[paramIndex] = externParam
-                    .WithType(IdentifierName(nameof(SafeHandle)).WithTrailingTrivia(TriviaList(Space)));
+                // Accept the SafeHandle instead. Use concrete safe handle type to help avoid cases when wrong safe handle is passed to the method
+                // unless it is a basic HANDLE, which is annotated in metadata to be closed with CloseHandle function but is widely used across
+                // Windows API in places where it is closed via some domain-specific function
+                TypeSyntax safeHandleSyntax = releaseMethod != "CloseHandle" && this.RequestSafeHandle(releaseMethod) is { } safeHandleType ?
+                    safeHandleType
+                    : IdentifierName(nameof(SafeHandle));
+                parameters[paramIndex] = externParam.WithType(safeHandleSyntax.WithTrailingTrivia(TriviaList(Space)));
 
                 // hParamNameLocal;
                 arguments[paramIndex] = Argument(typeDefHandleName);
--- a/test/CsWin32Generator.Tests/CsWin32GeneratorTests.cs
+++ b/test/CsWin32Generator.Tests/CsWin32GeneratorTests.cs
@@ -192,18 +192,18 @@ public partial class CsWin32GeneratorTests : CsWin32GeneratorTestsBase
         ["ID3D11DeviceChild", "GetPrivateData", "this winmdroot.Graphics.Direct3D11.ID3D11DeviceChild @this, in global::System.Guid guid, ref uint pDataSize, [Optional] Span<byte> pData"],
         ["WriteFile", "WriteFile", "SafeHandle hFile, in byte lpBuffer, uint* lpNumberOfBytesWritten, global::System.Threading.NativeOverlapped* lpOverlapped", false],
         // All params included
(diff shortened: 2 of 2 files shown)


mjr4077au:
There's at least 2-3 friendly overloads I've encountered that use a SafeFileHandle instead of a more correct type like SafeProcessHandle, SafeThreadHandle, or SafeAccessTokenHandle. I'm not saying this PR isn't for the better, but it could generate quite a lot of compilation errors for some people's projects.

DoctorKrolic:
> There's at least 2-3 friendly overloads I've encountered that use a SafeFileHandle instead of a more correct type like SafeProcessHandle, SafeThreadHandle, or SafeAccessTokenHandle.

This is due to the fact that basic `HANDLE` type is used widely across WinAPI. In most cases correct closing function for it is `CloseHandle`, but in some cases such handle is closed via some domain-specific function. In metadata this hanlde type is annotated with `CloseHandle` as its closing function, which corresponds to `SafeFileHandle` projection by CsWin32. I specificly excluded `HANDLE` from the logic of this PR for that reason, so in places where `HANDLE` is used CsWin32 will still generate base `SafeHandle` type to account for the case you've described. In any case, if you provide concrete examples of such broken functions, we can investigate and either fix them on metadata side or add to test suit to verify correct signature here

mjr4077au:
It sounds like I shouldn't have too much cause for concern. I do generate my own SafeHandle types for certain things though and it has been handy to be able to send them directly through to a friendly method overload. I suppose I'll see how things like when/if this is merged and address whatever's impacted.

DoctorKrolic:
> I do generate my own SafeHandle types for certain things though

Can you please describe your use cases in more details? My line of thoughts was: you already get a safe handle out of friendly overload of a function that creates it. Then you use it in other functions, which will accept the same handle type after this PR. Then the handle is automatically closed when disposed and this functionality is also already provided by the CsWin32. Why would you need to delcare your own safe handle type then? I mean yes, in cases of unannotated `HANDLE` the creating function will wrongly return `SafeFileHandle`, so you might want to delcare you own safe handle type and use it instead for this case. But as I said above, such cases are not affected by this PR, so we should be good

jevansaks:
What is the issue that this PR is fixing? With no updated or added test collateral in the change it's hard for me to understand what the effective result is.

Is the idea that the parameters should require the more specific type to help developers know what types to pass in? I agree with @mjr4077au that this might have negative impact for folks where the type hasn't been so strict thus far. Perhaps there's a better way to document the _recommended_ parameter type without requiring it?

mjr4077au:
A prime example would be a `SafePinnedGCHandle` type I've made that looks like the below. I can then use this class with a binary form SID stored as `byte[]` with any friendly overload that requires a SID be provided as a `SafeHandle`.
```C#
using System.Runtime.InteropServices;
using PSADT.LibraryInterfaces.SafeHandles;

namespace PSADT.SafeHandles
{
    /// <summary>
    /// Represents a wrapper for a <see cref="GCHandle"/> that pins an object in memory.
    /// </summary>
    internal sealed class SafePinnedGCHandle : SafeMemoryHandle<SafePinnedGCHandle>
    {
        /// <summary>
        /// Allocates a new <see cref="SafePinnedGCHandle"/> for the specified object, pinning it in memory.
        /// </summary>
        /// <param name="value"></param>
        /// <returns></returns>
        internal static SafePinnedGCHandle Alloc<T>(T[] value)
        {
            return new(GCHandle.Alloc(value, GCHandleType.Pinned), Marshal.SizeOf<T>() * value.Length, true);
        }

        /// <summary>
        /// Constructs a new instance of the <see cref="SafePinnedGCHandle"/> class with the specified <see cref="GCHandle"/>.
        /// </summary>
        /// <param name="handle"></param>
        /// <param name="length"></param>
        /// <param name="ownsHandle"></param>
        private SafePinnedGCHandle(GCHandle handle, int length, bool ownsHandle) : base(handle.AddrOfPinnedObject(), length, ownsHandle)
        {
            pinnedHandle = handle;
        }

        /// <summary>
        /// Releases the handle by freeing the underlying <see cref="GCHandle"/>.
        /// </summary>
        /// <returns></returns>
        protected override bool ReleaseHandle()
        {
            pinnedHandle.Free();
            return true;
        }

        /// <summary>
        /// The underlying <see cref="GCHandle"/> that holds the pinned object.
        /// </summary>
        private readonly GCHandle pinnedHandle;
    }
}
```
If the friendly overloads change from `SafeHandle` to `FreeSidSafeHandle` or `LocalFreeSafeHandle`, then I'd have to still construct a GCHandle to pin the array or use a fixed statement, then construct eiher one of those CsWin32 SafeHandle types myself and be absolutely certain I don't give it ownership of the pinned address I'm passing through, which could have dangerous repercussions during runtime.

mjr4077au:
I just ran into a situation where I had a a method set up to use `SafeFileHandle` because that's what `GetCurrentProcess_SafeHandle()` returns, but then I couldn't use `System.Diagnostics.Process`'s `SafeHandle` member with it because it (rightly) gives you a `SafeProcessHandle`. Unless CsWin32 can be better about the SafeHandle types it uses (using `SafeProcessHandle` where it makes sense to, etc), I think this PR will cause quite a few interop issues like this.

PR #1668, opened 2026-04-01
Bump metadata version

DoctorKrolic:
Bumps metadata version. It took ober an hour on my machine, but I swear I ran all the tests including the "full generation" ones, added a little test fix and all tests where green

### DIFF
--- a/test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
+++ b/test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
@@ -247,11 +247,7 @@ using System.Runtime.CompilerServices;
                 }
             }
 
-            // Filter out SYSLIB1092 diagnostics (related to DisableRuntimeMarshalling) as they are expected
-            var filteredGeneratorDiagnostics = generatorDiagnostics.Where(d =>
-                !(d.Descriptor.Id == "SYSLIB1092" && d.Descriptor.Title.ToString().Contains("The return value in the managed definition will be converted to an additional 'out' parameter at the end of the parameter list when calling the unmanaged COM method.", StringComparison.OrdinalIgnoreCase)));
-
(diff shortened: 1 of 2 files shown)


test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
```diff
@@ -247,11 +247,7 @@ protected async Task CompileGeneratedFilesWithSourceGenerators(string outputPath
                 }
             }
 
-            // Filter out SYSLIB1092 diagnostics (related to DisableRuntimeMarshalling) as they are expected
-            var filteredGeneratorDiagnostics = generatorDiagnostics.Where(d =>
-                !(d.Descriptor.Id == "SYSLIB1092" && d.Descriptor.Title.ToString().Contains("The return value in the managed definition will be converted to an additional 'out' parameter at the end of the parameter list when calling the unmanaged COM method.", StringComparison.OrdinalIgnoreCase)));
```
DoctorKrolic, line 252: English is not my main language, so due to the message check I first got a bunch of red tests because of to this diagnostic. The message check isn't necessary here since diagnostic ID on its own is its unique identifier, but I figured even a better fix: this diagnostic is reported in our generated files, which have suppressions for it on the top. So it is better to filter out suppressed diagnostics in general since it is more representative of a real-world usage

DoctorKrolic:
@jevansaks PTAL

test/CsWin32Generator.Tests/CsWin32GeneratorTestsBase.cs
jevansaks: Yes I like that idea! Thanks, I just did not know about that flag.

jevansaks (approved):

DoctorKrolic:
CI seems to be stuck, can you restart it please?

PR #1683, opened 2026-04-25
Skip _unmanaged emission when struct layout matches _managed variant

### LINKED ISSUE #1682: CreateBitmapFromDxgiSurface uses D2D1_BITMAP_PROPERTIES1_unmanaged even though D2D1_BITMAP_PROPERTIES1 already has the   same layout
## Actual behavior

When generating bindings for `ID2D1DeviceContext6`, CsWin32 generates `ID2D1DeviceContext.CreateBitmapFromDxgiSurface` with `D2D1_BITMAP_PROPERTIES1_unmanaged` instead of `D2D1_BITMAP_PROPERTIES1`.

The generated signature is:

```csharp
unsafe void CreateBitmapFromDxgiSurface(
    IDXGISurface surface,
    [Optional] D2D1_BITMAP_PROPERTIES1_unmanaged* bitmapProperties,
    out ID2D1Bitmap1 bitmap);
```

CsWin32 also generates an extra `D2D1_BITMAP_PROPERTIES1_unmanaged` struct.

In the generated output, `D2D1_BITMAP_PROPERTIES1_unmanaged` appears to have the same layout and field types as `D2D1_BITMAP_PROPERTIES1`, including `ID2D1ColorContext_unmanaged* colorContext`.

Because of this, I have to construct `D2D1_BITMAP_PROPERTIES1_unmanaged` in my own code to call `CreateBitmapFromDxgiSurface`, even though the Win32 API uses `const D2D1_BITMAP_PROPERTIES1*` and `D2D1_BITMAP_PROPERTIES1` already seems usable as-is.

## Expected behavior

`CreateBitmapFromDxgiSurface` should accept `D2D1_BITMAP_PROPERTIES1` (or `D2D1_BITMAP_PROPERTIES1?` in the convenience overload), and CsWin32 should not generate a duplicate `D2D1_BITMAP_PROPERTIES1_unmanaged` type for this case.

## Repro steps

```txt
ID2D1DeviceContext6
```

```json
{
  "$schema": "https://aka.ms/CsWin32.schema.json"
}
```

```csharp
var bitmapProperties = new D2D1_BITMAP_PROPERTIES1_unmanaged() {
    bitmapOptions = D2D1_BITMAP_OPTIONS_TARGET | D2D1_BITMAP_OPTIONS_CANNOT_DRAW,
    pixelFormat = new() {
        format = DXGI_FORMAT_UNKNOWN,
        alphaMode = D2D1_ALPHA_MODE_PREMULTIPLIED
    },
    dpiX = _dpi,
    dpiY = _dpi,
    colorContext = null
};

_d2dDeviceContext.CreateBitmapFromDxgiSurface(tempDXGISurface, bitmapProperties, out _d2dRenderTarget[i]);
```

### Context

- CsWin32 version: `0.3.275`
- Win32Metadata version (if explicitly set by project): not explicitly set
- Target Framework: `net11.0-windows10.0.17763.0`
- `LangVersion` (if explicitly set by project): not explicitly set
- `DisableRuntimeMarshalling`: `true`
- `PublishAot`: `true`

_This issue report was drafted with help from Codex based on the generated bindings and local project configuration._

MukundaKatta:
Closes #1682.

`CreateBitmapFromDxgiSurface` was projected as `_unmanaged` even though the `_managed` variant had identical signature and marshaling, forcing callers into `unsafe` for no benefit. Added a layout-equality check before `_unmanaged` emission so structs that match the safe variant are routed there.

Test in `test/Microsoft.Windows.CsWin32.Tests/COMTests.cs`.

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -1219,6 +1219,11 @@ public partial class Generator : IGenerator, IDisposable
     /// <summary>
     /// Determines whether the given type definition needs an "_unmanaged" suffix in its projection.
     /// </summary>
+    /// <param name="reader">The metadata reader for the assembly containing the type.</param>
+    /// <param name="typeDef">The type definition to check.</param>
+    /// <param name="allowMarshaling">A value indicating whether marshaling is allowed.</param>
+    /// <param name="isManagedType">A value indicating whether the type is a managed type.</param>
+    /// <returns><see langword="true"/> if the type should have an "_unmanaged" suffix; otherwise <see langword="false"/>.</returns>
     /// <remarks>
     /// Under COM source generators, structs cannot have a marshaled projection that differs from the unmanaged one
     /// (managed pointer fields always demote to _unmanaged*), so the two variants would be byte-for-byte identical.
--- a/src/Microsoft.Windows.CsWin32/Generator.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.cs
@@ -1040,7 +1040,7 @@ public partial class Generator : IGenerator, IDisposable
                 }
             }
 
-            bool hasUnmanagedName = this.HasUnmanagedSuffix(this.Reader, typeDef.Name, context.AllowMarshaling, this.IsManagedType(typeDefHandle));
+            bool hasUnmanagedName = this.HasUnmanagedSuffix(this.Reader, typeDef, context.AllowMarshaling, this.IsManagedType(typeDefHandle));
             this.volatileCode.GenerateType(typeDefHandle, hasUnmanagedName, delegate
             {
                 if (this.RequestInteropTypeHelper(typeDefHandle, context) is MemberDeclarationSyntax typeDeclaration)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -1216,9 +1216,33 @@ internal void GetBaseTypeInfo(TypeDefinition typeDef, out StringHandle baseTypeN
...
+    /// Skip the suffix in that case so callers don't need a redundant struct type. See https://github.com/microsoft/CsWin32/issues/1682.
+    /// </remarks>
+    internal bool HasUnmanagedSuffix(MetadataReader reader, TypeDefinition typeDef, bool allowMarshaling, bool isManagedType)
+    {
+        if (!this.HasUnmanagedSuffix(reader, typeDef.Name, allowMarshaling, isManagedType))
+        {
+            return false;
+        }
+
+        // Source generators force every managed-pointer struct field through the unmanaged projection,
```
jevansaks, line 1239: What is the expected impact for built-in COM? I know I looked into this and the logic for when a struct needed to be managed vs unmanaged was fairly convoluted and had to do with whether there was a self-reference cycle detected in the structs and I didn't have time to keep digging in.

Is the goal to leave built-in COM alone and only try to improve this for the source-generated outputs?

MukundaKatta:
@jevansaks no impact on built-in COM. The new path is gated behind `useSourceGenerators` — `HasUnmanagedSuffix` returns `isInterface || !this.useSourceGenerators`, so when source generators are off the predicate keeps returning `true` exactly like before, and the existing managed/unmanaged self-reference logic still drives mangling. Only the source-generator path skips emitting the redundant `_unmanaged` clone of layout-equivalent structs. Happy to add a built-in COM regression test if you want belt-and-suspenders.

jevansaks:
Sorry for the delay, I just merged a PR to enable all the test cases running in PR so if you could merge in latest origin/main we can get that pipeline queued?

jevansaks:
@MukundaKatta I've updated your branch to latest and fixed the basic build errors. Now seeing test failures in the PR run. You should be able to see these and investigate them. Thanks!

PR #1693, opened 2026-05-07
Provide an additional friendly overload with raw unconvertible handles when necessary

DoctorKrolic:
Previously if friendly overload accepted a handle with a known release method CsWin32 would always "promote" it to a `SafeHandle`. This works well if CsWin32 can trivially represent that handle as a `SafeHandle` so that function that opens/creates the handle in the first place returns it a a custom generated `SafeHandle`. However, if release method is a complicated one and CsWin32 cannot generate a trivial `SafeHandle` wrapper based on it user was forced to declare such safe handle type on their own to use it in other friendly overloads. This might not be ideal e.g. if the handle is used whithin one method frame so that declaring a safe handle type for it is an overkill. Now CsWin32 generates another overload with such handles represented as "raw" handles so that user has a choice either to use the raw handle or to declare their own safe handle type and use it instead

### DIFF
--- a/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
+++ b/src/Microsoft.Windows.CsWin32/Generator.FriendlyOverloads.cs
@@ -90,7 +90,7 @@ public partial class Generator
 
         bool improvePointersToSpansAndRefs = this.canUseSpan;
         FriendlyMethodBookkeeping bookkeeping = new();
-        foreach (MethodDeclarationSyntax method in this.DeclareFriendlyOverload(methodDefinition, externMethodDeclaration, declaringTypeName, overloadOf, helperMethodsAdded, avoidWinmdRootAlias, improvePointersToSpansAndRefs, omitOptionalParams: false, bookkeeping))
+        foreach (MethodDeclarationSyntax method in this.DeclareFriendlyOverload(methodDefinition, externMethodDeclaration, declaringTypeName, overloadOf, helperMethodsAdded, avoidWinmdRootAlias, improvePointersToSpansAndRefs, omitOptionalParams: false, promoteUnconvertibleHandles: true, bookkeeping))
         {
             yield return method;
         }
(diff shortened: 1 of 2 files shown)


jevansaks:
There is already a safeHandles knob in NativeMethods.json. How is this different than letting the consumer choose a different form of the projection? More overloads can just be more confusing.

It sounds like from the description this is limited to "complex" handles, could you give some concrete examples for those and how this would help?

DoctorKrolic:
> There is already a safeHandles knob in NativeMethods.json. How is this different than letting the consumer choose a different form of the projection?

`useSafeHandles` switch disables _all_ safe handles, while this change targets a very specific use case. Some handles aren't trivially closable, i.e. their annotated close method takes 2 or more parameters. In such cases CsWin32 doesn't generate a safe hanlde implementation for them, so the function openning the handle just returns a raw handle instance. At the other hand functions which accept the handle still require a `SafeHandle` for that handle parameter. This means that a user _must_ declare their own safe handle wrapper to be able to use friendly overloads

> could you give some concrete examples for those and how this would help

The added test functions show this quite well. In order to enumerate WFP filters you need to open an enum handle via `FwpmFilterCreateEnumHandle0`. Since the returned handle is closed via `FwpmFilterDestroyEnumHandle0`, which takes 2 paramneters (session handle + enum handle), CsWin32 cannot generate a safe handle wrapper for it and returns it as `out FWPM_FILTER_ENUM_HANDLE`. To actually enumerate filters function `FwpmFilterEnum0` is used. It takes enum handle as its second parameter. Since it is a closable handle CsWin32 promotes it to a `SafeHandle`. So now as a user I get the handle as `FWPM_FILTER_ENUM_HANDLE`, but need a `SafeHandle` on a use site, so I _have to_ declare a safe handle wrapper myself. This might be an overkill. In my actuall use case I needed to enumerate filters until I find a permanent one on a given sublayer, so declaring my own safe handle just to wrap the value in it and dispose at the end of the method felt like unnecessary boilerplate. With this change 2 sets of overloads will be generated for such functions: in one all handles stay `SafeHandle`s and in the second one such 'unconvertable' handles are downgraded to the 'raw' handle types. This gives a user flexibility to either declare a safe handle and use it if the handle is long lived and automatic lifetime tracking is desired or just simply use the raw handle if their logic is simple enough

DoctorKrolic:
@jevansaks PTAL

mjr4077au:
DoctorKrolic, maybe don't ping and hassle maintainers? There's already not a lot of development occuring with this project, and being hassled is a sure fire way (at least for me) to put soemthing like this at the absolute bottom of my pile.

jevansaks:
I've noodled on this and I am not a fan of this approach. It's easy enough for callers to make a non-owning SafeHandle to pass in (e.g. `new SafeFileHandle(h, false)`) and there are other cases where callers have to do this too (i.e. pass "null" for some SafeHandle parameters).

If this is a hardship, can we talk about what it takes to project the methods which return raw handles as the appropriate generated SafeHandle type? IIRC from other threads the CloseHandle method takes multiple parameters so is the right thing for the generated code to somehow get the other parameters passed in or from some available context?

PR #1701, opened 2026-05-24
Add extensionReceiver option for layered PInvoke composition (#1477)

### LINKED ISSUE #1477: Allow specifying extension member target for static class

When composing multiple assemblies that use CsWin32 you need to have multiple definitions of the main static class. In WinForms this ends up being `PInvokeCore` and `PInvoke`. This makes discovery difficult.

Allow specifying an `extension` receiver class name that will result in the static class generating a specified extension scope.

In the `NativeMethods.json`:

```json
{
  "className": "WinFormsPInvokes",
  "extensionReciever": "PInvoke"
}
```

Generates (effectively):

```C#
namespace Windows.Win32;

internal partial static class WinFormsPInvokes
{
    extension (PInvoke)
    {
        // All defines generated here
    }
}
```

This feature requires C# 14 (with .NET 10) to compile but works on both .NET Framework and .NET (core). You can still reference the members directly via the static class name so consuming libraries can still use this without using C# 14. I don't think it is necessary to add any deduplication logic as part of this initial feature request.

Documentation:

https://learn.microsoft.com/dotnet/csharp/whats-new/csharp-14#extension-members
https://learn.microsoft.com/dotnet/csharp/language-reference/proposals/csharp-14.0/extensions

There aren't any other good options that I'm aware of.

**Additional context**

JeremyKuhne:
Implements `extensionReceiver` in `NativeMethods.json` so multiple assemblies running CsWin32 can contribute to a single shared static class (typically `PInvoke`). Callers reach every generated native API through one discovery symbol regardless of which assembly emitted it.

Closes #1477.

## What this adds

A new `extensionReceiver` setting in `NativeMethods.json` names a static class declared in another assembly that also runs CsWin32 to receive this assembly's generated members. When set, members are emitted inside a C# 14 `extension(<Receiver>) { ... }` block on the configured `className`. The receiver must live in the same namespace as `className`, be a static class, be accessible, and differ from `className`.

Constants which the C# 14 extension grammar cannot host are emitted as fields on the host class (preserving the configured visibility) and re-exposed through a forwarder property inside the extension block, so both `PInvoke.X` (extension path, runtime contexts) and `<HostClass>.X` (const path, enum initializers / attribute args / fixed-array sizes / case labels) work.

Cross-assembly duplicates are skipped using a full-signature match (name + arity + per-parameter type-name comparison after normalization that strips `global::` and leading namespace segments). User-authored wrappers with the same name but a different signature do not suppress the underlying metadata extern.

When `SuperGenerator` runs multiple generators (e.g. `Windows.Win32` and `Windows.Wdk` side by side), each generator silently no-ops the wrap if the receiver doesn't resolve in its own namespace; `PInvoke011` is reported only when no generator could resolve.

The analyzer is multi-targeted into a Roslyn 4.11 floor leg (feature disabled) and a Roslyn 5.0+ leg (feature enabled, using `ExtensionBlockDeclarationSyntax` / `INamedTypeSymbol.IsExtension`). NuGet picks the appropriate leg from the host's Roslyn version.

## Diagnostics

| ID | Cause |
|---|---|
| `PInvoke011` | `extensionReceiver` couldn't be resolved, isn't a static class, isn't accessible, or equals `className`. |
| `PInvoke012` | Consuming project requires C# 14 (default on .NET 10+). |
| `PInvoke013` | Host analyzer doesn't support C# 14 extension members. |

## Documentation

`docfx/docs/composition.md` is added to the docs TOC with the conceptual model, configuration reference, dispatch rules, multi-assembly composition behavior, a step-by-step migration playbook for existing multi-layer projects, common-issue catalog, and the diagnostics matrix. `features.md` and `getting-started.md` cross-link to it.

## Validation

* 16 new unit tests in `ExtensionReceiverTests.cs` covering shape, multi-file emission, gating, the four `TryResolveExtensionReceiver` outcomes, and the full-signature dedup matrix (name match, arity mismatch, parameter-type mismatch, generic user-wrapper vs. metadata extern of the same simple name).
* Synthetic integration test `integration-tests/sdk-extension` consuming the locally-packed package via the SDK 11 preview.
* Independently vetted in `dotnet/winforms`: https://github.com/JeremyKuhne/winforms/commit/45562f1db39533e5b1357c73230375973ce8f61c

Feedback from the WinForms vet drove three refinements over the initial design:

1. Dedup upgraded from name-only to full-signature match, eliminating a self-recursion hazard where a user-authored wrapper could silently suppress the metadata extern of the same simple name.
2. Constant backing fields keep their configured visibility instead of being forced to `private`, so they remain usable in C# const contexts.
3. Receiver resolution short-circuits per generator under `SuperGenerator`, removing false `PInvoke011` reports when one of several side-by-side generators legitimately can't see the receiver.

### DIFF
--- a/Directory.Packages.Analyzers.props
+++ b/Directory.Packages.Analyzers.props
@@ -36,6 +36,6 @@
     <PackageVersion Update="System.Memory" Version="4.5.5" />
     <PackageVersion Update="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
-    <PackageVersion Update="System.Text.Json" Version="8.0.0" />
+    <PackageVersion Update="System.Text.Json" Version="8.0.5" />
   </ItemGroup>
 </Project>
--- a/Directory.Packages.Analyzers.props
+++ b/Directory.Packages.Analyzers.props
@@ -13,10 +13,29 @@
     <PackageVersion Update="Microsoft.CodeAnalysis.Common" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp" Version="$(CodeAnalysisVersionForAnalyzers)" />
-    <PackageVersion Update="System.Collections.Immutable" Version="9.0.0" />
-    <PackageVersion Update="System.Memory" Version="4.6.0" />
+
+    <!--
+      ANALYZER SUPPORT-ASSEMBLY VERSION BINDINGS — DO NOT BUMP HERE.
+
+      The following System.* versions are the support assemblies that get copied next to
+      the analyzer (CopyLocalLockFileAssemblies=true) and shipped inside the NuGet package.
+      They are intentionally pinned to the FLOOR (Roslyn 4.11 / VS 2022 Update 14 / .NET 8)
+      versions because the analyzer must load inside older Visual Studio and `dotnet build`
+      hosts that ship exactly these assembly versions in the compiler's load context.
+      Shipping a NEWER version than the host provides risks an assembly-load/version
+      mismatch at analysis time (the root cause behind issue #1555).
+
+      These are the central defaults consumed by the Roslyn 4.11 floor leg
+      (Microsoft.Windows.CsWin32.Roslyn4). The Roslyn 5.0 leg
+      (Microsoft.Windows.CsWin32) runs in a newer host whose compiler genuinely ships
+      newer System.* assemblies, so that leg — and ONLY that leg — overrides these
+      versions locally via <VersionOverride> in its .csproj. Keep the floor numbers here
+      and add per-leg overrides there; never raise the floor to satisfy the newer leg.
+    -->
+    <PackageVersion Update="System.Collections.Immutable" Version="8.0.0" />
+    <PackageVersion Update="System.Memory" Version="4.5.5" />
     <PackageVersion Update="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
-    <PackageVersion Update="System.Text.Json" Version="8.0.5" />
+    <PackageVersion Update="System.Text.Json" Version="8.0.0" />
   </ItemGroup>
 </Project>
--- a/Directory.Packages.Analyzers.props
+++ b/Directory.Packages.Analyzers.props
@@ -1,14 +1,20 @@
 <Project>
   <PropertyGroup>
-    <CodeAnalysisVersionForAnalyzers>4.11.0</CodeAnalysisVersionForAnalyzers>
+    <!--
+      Default analyzer Roslyn version. The analyzer is multi-targeted so it ships an assembly
+      built against Roslyn 4.11 (the floor that supports VS 2022 Update 14 / .NET 8 SDK) AND
+      an assembly built against Roslyn 5.0 (required for C# 14 ExtensionBlockDeclarationSyntax).
+      Projects override $(CodeAnalysisVersionForAnalyzers) per Roslyn leg.
+    -->
+    <CodeAnalysisVersionForAnalyzers Condition="'$(CodeAnalysisVersionForAnalyzers)'==''">4.11.0</CodeAnalysisVersionForAnalyzers>
   </PropertyGroup>
   <ItemGroup>
     <!-- These versions carefully chosen to support VS 2022 Update 14 and also run down to .NET 8. -->
     <PackageVersion Update="Microsoft.CodeAnalysis.Common" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp" Version="$(CodeAnalysisVersionForAnalyzers)" />
-    <PackageVersion Update="System.Collections.Immutable" Version="8.0.0" />
-    <PackageVersion Update="System.Memory" Version="4.5.5" />
+    <PackageVersion Update="System.Collections.Immutable" Version="9.0.0" />
+    <PackageVersion Update="System.Memory" Version="4.6.0" />
     <PackageVersion Update="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
     <PackageVersion Update="System.Text.Json" Version="8.0.5" />
--- a/docfx/docs/composition.md
+++ b/docfx/docs/composition.md
@@ -1,0 +1,438 @@
+# Layered composition with `extensionReceiver`
+
+CsWin32 lets multiple assemblies contribute to a single, unified static class for native API
+discovery. One assembly hosts the canonical `PInvoke` symbol; every other assembly that uses
+CsWin32 extends it. Callers reach every native API through `PInvoke.X()` regardless of which
+assembly emitted it.
+
+This is built on C# 14 [extension members][csharp14-extensions] and is configured via the
+`extensionReceiver` setting in `NativeMethods.json`.
+
+## When to use it
+
+Use `extensionReceiver` if your codebase is layered — a low-level library, a UI/runtime helper
+library on top, and an application on top of that — and you want callers to use a single
+`PInvoke.X()` discovery surface regardless of which layer happens to declare `X`.
+
+Don't use it for single-assembly projects. The default `partial class PInvoke` already gives you
+everything you need without C# 14.
+
+## Requirements
+
+C# 14 or later in every project that consumes the generated extension members. This is the
+default when targeting .NET 10 or later; otherwise set `<LangVersion>14</LangVersion>` (or
+`Preview` / `Latest` / `LatestMajor`) in the project file.
+
+## Two roles
+
+Every assembly that runs CsWin32 plays exactly one role:
+
+| Role | `NativeMethods.json` shape | What the generator emits |
+|---|---|---|
+| **Owner** | `extensionReceiver` *not* set | A plain `partial class <className>` with members declared directly on it. |
+| **Extender** | `extensionReceiver: "<OwnerClassName>"` set | A `partial class <className>` whose members live inside an `extension(<OwnerClassName>) { … }` block. |
+
+In a layered stack, the lowest assembly is the owner; every higher assembly is an extender. Each
+extender must use a `className` that differs from the owner's and from every other extender's.
+
+## A worked example
+
+Three projects in dependency order: `MyApp.Core` (lowest), `MyApp.Helpers`, `MyApp.App` (highest).
+
+### Owner — `MyApp.Core`
+
+```jsonc
+// NativeMethods.json
+{
+  "$schema": "https://aka.ms/CsWin32.schema.json",
+  "className": "PInvoke",
+  "public": true
+}
+```
+
+```text
+// NativeMethods.txt
+GetTickCount
+```
+
+Generates:
+
+```csharp
+namespace Windows.Win32
+{
+    public static partial class PInvoke
+    {
+        [DllImport("KERNEL32.dll", ExactSpelling = true)]
+        public static extern uint GetTickCount();
+    }
+}
+```
+
+### Extender — `MyApp.Helpers` (references `MyApp.Core`)
+
+```jsonc
+// NativeMethods.json
+{
+  "$schema": "https://aka.ms/CsWin32.schema.json",
+  "className": "PInvokeHelpers",
+  "extensionReceiver": "PInvoke",
+  "public": true
+}
+```
+
+```text
+// NativeMethods.txt
+GetForegroundWindow
+```
+
+Generates:
+
+```csharp
+namespace Windows.Win32
+{
+    public static partial class PInvokeHelpers
+    {
+        extension (global::Windows.Win32.PInvoke)
+        {
+            [DllImport("USER32.dll", ExactSpelling = true)]
+            public static extern global::Windows.Win32.Foundation.HWND GetForegroundWindow();
+        }
+    }
+}
+```
+
+### Extender — `MyApp.App` (references both)
+
+```jsonc
+{
+  "$schema": "https://aka.ms/CsWin32.schema.json",
+  "className": "PInvokeApp",
+  "extensionReceiver": "PInvoke"
+}
+```
+
+```text
+IsWindow
+```
+
+### Consuming code
+
+Application code reaches every API through the single `PInvoke` symbol:
+
+```csharp
+using Windows.Win32;
+
+uint   ticks = PInvoke.GetTickCount();        // declared in MyApp.Core
+HWND   hwnd  = PInvoke.GetForegroundWindow(); // declared in MyApp.Helpers
+bool   alive = PInvoke.IsWindow(hwnd);        // declared in MyApp.App
+```
+
+The `using Windows.Win32;` directive at the top of the file is what makes the extension members
+discoverable. See [Dispatch rules](#dispatch-rules) below.
+
+## Dispatch rules
+
+C# 14 extension members participate in member lookup only when the static class that declares the
+`extension(T) { … }` block is in scope at the call site. Three practical rules cover almost every
+case.
+
+### Rule 1 — Bring the extender's namespace into scope
+
+To call `PInvoke.X()` and have the compiler find `X` in an extender, the call site must either be
+inside `namespace Windows.Win32 { … }` or contain a **file-scoped** `using Windows.Win32;`.
+
+> [!IMPORTANT]
+> `global using Windows.Win32;` in a `GlobalUsings.cs` file works for most lookups, but it does
+> **not** contribute to type-name resolution inside user-authored `extension(T)` blocks. If you
+> author your own extension blocks, add the file-scoped `using` in those files even if the same
+> namespace is declared as a global using.
+
+### Rule 2 — When you can't use the extension path, use the host class directly
+
+Some C# contexts cannot resolve through an extension property. Specifically:
+
+- `enum` initializers
+- attribute arguments
+- `fixed`-array sizes
+- `case` labels of `switch`
+
+For these contexts, reach the value through the host class:
+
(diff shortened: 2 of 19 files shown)


docfx/docs/composition.md
```diff
@@ -0,0 +1,438 @@
...
+This is built on C# 14 [extension members][csharp14-extensions] and is configured via the
+`extensionReceiver` setting in `NativeMethods.json`.
+
+## When to use it
+
+Use `extensionReceiver` if your codebase is layered — a low-level library, a UI/runtime helper
+library on top, and an application on top of that — and you want callers to use a single
+`PInvoke.X()` discovery surface regardless of which layer happens to declare `X`.
+
+Don't use it for single-assembly projects. The default `partial class PInvoke` already gives you
```
jevansaks, line 17: But would it be wrong? Why provide a choice? In C# 14 shouldn't we _always_ do this?

docfx/docs/composition.md
```diff
@@ -0,0 +1,438 @@
...
+
+> [!IMPORTANT]
+> `global using Windows.Win32;` in a `GlobalUsings.cs` file works for most lookups, but it does
+> **not** contribute to type-name resolution inside user-authored `extension(T)` blocks. If you
+> author your own extension blocks, add the file-scoped `using` in those files even if the same
+> namespace is declared as a global using.
+
+### Rule 2 — When you can't use the extension path, use the host class directly
+
+Some C# contexts cannot resolve through an extension property. Specifically:
```
jevansaks, line 152: Are these bugs that you've filed on roslyn? Please track them here

docfx/docs/composition.md
```diff
@@ -0,0 +1,438 @@
...
+```
+
+Setting `public: true` on the owner makes the receiver visible to any consumer. If your owner
+intends to keep all its API surface internal, you can instead apply
+`[InternalsVisibleTo("<each extender>")]` from the owner — but be aware that final application code
+won't be able to use `PInvoke.X()` to reach extender members unless it, too, can see the owner's
+receiver type.
+
+> [!NOTE]
+> Setting `public: true` on the owner publishes **every** type CsWin32 emits in that assembly,
```
jevansaks, line 294: I haven't looked at the winforms consumption yet but this seems problematic if we start having parts of the ecosystem shipping these things as public.

docfx/docs/getting-started.md
```diff
@@ -37,25 +37,25 @@ dotnet add package System.Memory
 dotnet add package System.Runtime.CompilerServices.Unsafe
 ```
 
-Projects targeting .NET do *not* need to add these package references, although it is harmless to do so.
+Projects targeting .NET do _not_ need to add these package references, although it is harmless to do so.
```
jevansaks: underscores are italics, asterisk is bold. this is changing the formatting unnecessarily.

Directory.Packages.Analyzers.props
```diff
@@ -1,14 +1,20 @@
...
+    <CodeAnalysisVersionForAnalyzers Condition="'$(CodeAnalysisVersionForAnalyzers)'==''">4.11.0</CodeAnalysisVersionForAnalyzers>
   </PropertyGroup>
   <ItemGroup>
     <!-- These versions carefully chosen to support VS 2022 Update 14 and also run down to .NET 8. -->
     <PackageVersion Update="Microsoft.CodeAnalysis.Common" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="$(CodeAnalysisVersionForAnalyzers)" />
     <PackageVersion Update="Microsoft.CodeAnalysis.CSharp" Version="$(CodeAnalysisVersionForAnalyzers)" />
-    <PackageVersion Update="System.Collections.Immutable" Version="8.0.0" />
-    <PackageVersion Update="System.Memory" Version="4.5.5" />
+    <PackageVersion Update="System.Collections.Immutable" Version="9.0.0" />
```
jevansaks: We've gotten in a lot of trouble changing these versions in the past see https://github.com/microsoft/CsWin32/pull/1555

src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec
```diff
@@ -43,13 +43,32 @@
...
+    <!-- ============================================================ -->
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Windows.SDK.Win32Docs.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Windows.SDK.Win32Docs.dll" />
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Windows.CsWin32.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Windows.CsWin32.dll" />
+    <file src="$SignedOutputPath$MessagePack.dll" target="analyzers\dotnet\roslyn4.11\cs\MessagePack.dll" />
+    <file src="$SignedOutputPath$MessagePack.Annotations.dll" target="analyzers\dotnet\roslyn4.11\cs\MessagePack.Annotations.dll" />
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Bcl.AsyncInterfaces.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Bcl.AsyncInterfaces.dll" />
+    <file src="$Roslyn4BaseOutputPath$System.Text.Json.dll" target="analyzers\dotnet\roslyn4.11\cs\System.Text.Json.dll" />
+    <file src="$Roslyn4BaseOutputPath$System.Text.Encodings.Web.dll" target="analyzers\dotnet\roslyn4.11\cs\System.Text.Encodings.Web.dll" />
+
     <file src="$CsWin32GeneratorBasePath$**" target="build\tools\" exclude="**\*.exe;**\MessagePack*.dll" />
```
jevansaks, line 72: Instead of multi-targeting the source generator we can also just have the new mode be part of the tool-based generator only.

docfx/docs/composition.md
JeremyKuhne: @jevansaks It is a pretty big breaking change to those that already have multiple projects with CsWin32 visibility. I suppose we could make it an opt-out and default to extending `PInvoke` if you haven't customized the class name?

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -1405,6 +1407,355 @@ private IReadOnlyList<ISymbol> FindTypeSymbolsIfAlreadyAvailable(string fullyQua
...
+    /// <summary>
+    /// Returns the cached, resolved extension receiver symbol or <see langword="null"/> when the option is unset or invalid.
+    /// </summary>
+    /// <returns>The receiver named type symbol, or <see langword="null"/>.</returns>
+    private INamedTypeSymbol? GetExtensionReceiverSymbol()
+    {
+        if (!this.extensionReceiverSymbolResolved)
+        {
+            this.extensionReceiverSymbolResolved = true;
+            this.TryResolveExtensionReceiver(out this.extensionReceiverSymbolCache, out _);
```
jevansaks, line 1516: Are we ignoring the error because we assume we've already checked for the extension receiver and it's going to succeed?

src/Microsoft.Windows.CsWin32/Generator.cs
```diff
@@ -1405,6 +1407,355 @@ private IReadOnlyList<ISymbol> FindTypeSymbolsIfAlreadyAvailable(string fullyQua
...
+    /// </summary>
+    /// <param name="memberName">The simple member name to look for.</param>
+    /// <param name="parameterTypeNames">The CsWin32-projected textual form of each parameter type, in declaration order. Pass an empty array when probing for a field/property/parameterless member (e.g. a constant). The strings are compared after a normalization that strips <c>global::</c> prefixes and leading namespace segments, so callers may pass fully-qualified names produced by <c>ToTypeSyntax(...)</c>.</param>
+    /// <returns><see langword="true"/> if a matching extension member is already declared on the receiver in scope; <see langword="false"/> otherwise (including when no extension receiver is configured or running on the Roslyn 4 leg).</returns>
+#pragma warning disable SA1202 // Keep dedup helper next to TryResolveExtensionReceiver / receiver-related code.
+    internal bool IsExtensionMemberAlreadyOnReceiver(string memberName, IReadOnlyList<string> parameterTypeNames)
```
jevansaks: What about when there is a property/constant versus parameterless method of the same name? Should we pass in the member kind too?

jevansaks (commented):
I'm about 80% done but I'll leave these comments so far. Will try to come back to it tomorrow.

Overall looking good. One thing I would love to see in this mode is that all the friendly overloads that we currently put in another namespace (and thus are totally undiscoverable and sometimes people miss them because they forget to include the Windows.Win32 namespace itself) -- can we use this extension everything approach to put them just in the main namespace or even on the main class as extension methods of that class?

docfx/docs/composition.md
JeremyKuhne: Not a bug, just a constraint.

docfx/docs/composition.md
JeremyKuhne: We're not making anything implicitly public as part of this, it is just a note on behavior.

src/Microsoft.Windows.CsWin32/Generator.cs
JeremyKuhne: Adding a comment

src/Microsoft.Windows.CsWin32/Generator.cs
JeremyKuhne: Good catch, captured

Directory.Packages.Analyzers.props
JeremyKuhne: I know, now the other project is pinned at it's current version.

docfx/docs/getting-started.md
JeremyKuhne: Removed

src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec
JeremyKuhne: We would still need to build against the new types, so I'm not sure how this helps?

JeremyKuhne:
> I'm about 80% done but I'll leave these comments so far. Will try to come back to it tomorrow.
> 
> Overall looking good. One thing I would love to see in this mode is that all the friendly overloads that we currently put in another namespace (and thus are totally undiscoverable and sometimes people miss them because they forget to include the Windows.Win32 namespace itself) -- can we use this extension everything approach to put them just in the main namespace or even on the main class as extension methods of that class?

Maybe, I'll look into it.

JeremyKuhne:
> > I'm about 80% done but I'll leave these comments so far. Will try to come back to it tomorrow.
> > Overall looking good. One thing I would love to see in this mode is that all the friendly overloads that we currently put in another namespace (and thus are totally undiscoverable and sometimes people miss them because they forget to include the Windows.Win32 namespace itself) -- can we use this extension everything approach to put them just in the main namespace or even on the main class as extension methods of that class?
> 
> Maybe, I'll look into it.

@jevansaks Can you give some concrete examples? I'm happy to lean into this pattern and continue to follow up where we think it would be useful.

Directory.Packages.Analyzers.props
JeremyKuhne: I made this much more in your face with the overrides coming downstream.

docfx/docs/getting-started.md
```diff
@@ -37,25 +37,25 @@ dotnet add package System.Memory
...
 Use of the `unsafe` keyword is required anywhere you use pointers.
 The source generator NuGet package sets the default value of the `AllowUnsafeBlocks` property for your project to `true`,
 but if you explicitly set it to `false` in your project file, generated code may produce compiler errors.
 
 Create a `NativeMethods.txt` file in your project directory that lists the APIs to generate code for.
-Each line may consist of *one* of the following:
+Each line may consist of one of the following:
 
-* Exported method name (e.g. `CreateFile`). This *may* include the `A` or `W` suffix, where applicable. This *may* be qualified with a namespace but is only recommended in cases of ambiguity, which CsWin32 will prompt where appropriate.
+* Exported method name (e.g. `CreateFile`). This may include the `A` or `W` suffix, where applicable. This may be qualified with a namespace but is only recommended in cases of ambiguity, which CsWin32 will prompt where appropriate.
```
jevansaks: Why are we removing all the emphasis from this doc? Can you put this back, or was there a reason you removed it?

docfx/docs/getting-started.md
```diff
@@ -136,16 +137,16 @@ Then you have two options:
...
 
 ```xml
 ```
 
 It is also strongly recommended to set DisableRuntimeMarshalling=true as shown here to ensure that the COM source generators handle
-all the marshalling correctly. See https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation and
-https://learn.microsoft.com/en-us/dotnet/standard/native-interop/comwrappers-source-generation for more information on these .NET source
+all the marshalling correctly. See <https://learn.microsoft.com/en-us/dotnet/standard/native-interop/pinvoke-source-generation> and
```
jevansaks: What do angle brackets on links do?

src/Microsoft.Windows.CsWin32/Generator.Constant.cs
```diff
@@ -420,7 +435,68 @@ private FieldDeclarationSyntax DeclareConstant(FieldDefinition fieldDef)
 
     private ClassDeclarationSyntax DeclareConstantDefiningClass()
     {
+#if ROSLYN5
+        // Skip the field-split when no receiver is configured OR when THIS generator's namespace can't
+        // resolve the configured receiver (multi-namespace pipeline edge case — see Generator.WrapAsExtensionMembers).
```
jevansaks, line 440: Since this talks about the multi-namespace case, how _does_ this work for multiple cswin32 namespaces? Today if you have Win32 and WDK then you get 2 pinvoke classes right so how do we decide which one to add it to?

src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec
jevansaks: Fair enough.

Directory.Packages.Analyzers.props
```diff
@@ -1,16 +1,41 @@
...
+      newer System.* assemblies, so that leg — and ONLY that leg — overrides these
+      versions locally via <VersionOverride> in its .csproj. Keep the floor numbers here
+      and add per-leg overrides there; never raise the floor to satisfy the newer leg.
+    -->
     <PackageVersion Update="System.Collections.Immutable" Version="8.0.0" />
     <PackageVersion Update="System.Memory" Version="4.5.5" />
     <PackageVersion Update="System.Reflection.Metadata" Version="8.0.0" />
     <PackageVersion Include="System.Text.Encodings.Web" Version="8.0.0" />
-    <PackageVersion Update="System.Text.Json" Version="8.0.5" />
+    <PackageVersion Update="System.Text.Json" Version="8.0.0" />
```
jevansaks: I think this one went back too far

src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec
```diff
@@ -43,13 +43,32 @@
...
+    <!-- The `extensionReceiver` feature is rejected on this leg via diagnostic PInvoke013. -->
+    <!-- ============================================================ -->
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Windows.SDK.Win32Docs.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Windows.SDK.Win32Docs.dll" />
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Windows.CsWin32.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Windows.CsWin32.dll" />
+    <file src="$SignedOutputPath$MessagePack.dll" target="analyzers\dotnet\roslyn4.11\cs\MessagePack.dll" />
+    <file src="$SignedOutputPath$MessagePack.Annotations.dll" target="analyzers\dotnet\roslyn4.11\cs\MessagePack.Annotations.dll" />
+    <file src="$Roslyn4BaseOutputPath$Microsoft.Bcl.AsyncInterfaces.dll" target="analyzers\dotnet\roslyn4.11\cs\Microsoft.Bcl.AsyncInterfaces.dll" />
+    <file src="$Roslyn4BaseOutputPath$System.Text.Json.dll" target="analyzers\dotnet\roslyn4.11\cs\System.Text.Json.dll" />
+    <file src="$Roslyn4BaseOutputPath$System.Text.Encodings.Web.dll" target="analyzers\dotnet\roslyn4.11\cs\System.Text.Encodings.Web.dll" />
```
jevansaks, line 70: Is this "roslynXXX" directory structure understood and automatically handled by VS and all other consumers of analyzers?

jevansaks (changes requested):

docfx/docs/getting-started.md
JeremyKuhne: Forces it to render as a link, GitHub makes it clickable without it so it isn't really necessary.

docfx/docs/getting-started.md
JeremyKuhne: I'll put it back, just thought it wasn't that useful.

src/Microsoft.Windows.CsWin32/Generator.Constant.cs
JeremyKuhne: A member always lands on the receiver in the same namespace as the generator that produced it, so the two PInvoke classes compose independently.

Directory.Packages.Analyzers.props
JeremyKuhne: You're right, good catch.

src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec
JeremyKuhne: Yes.

docfx/docs/getting-started.md
jevansaks: I'm still seeing unnecessary formatting changes in this delta, can you please revert?

docfx/docs/getting-started.md
jevansaks: And these are processed through DocFx over to https://microsoft.github.io/CsWin32/docs/getting-started.html where it's already working so ... just leave it please?

JeremyKuhne: Ugh, sorry, `markdownlint` got over aggressive. :)

jevansaks (approved):

Issue #1704, opened 2026-05-27
`IComIID` is not emitted on `net472` / `netstandard2.0`


JeremyKuhne:
## Summary

When a project multi-targets `net10.0` and `net472` (or `netstandard2.0`) and references [`Microsoft.Windows.CsWin32`](https://www.nuget.org/packages/Microsoft.Windows.CsWin32),
the generator emits the `IComIID` interface and attaches it to every generated COM struct **only on the .NET 7+ TFMs**. On `net472`:

1. The `Windows.Win32.IComIID` interface itself is not generated.
2. Generated COM struct wrappers (e.g. `IRunningObjectTable`, `IMoniker`, `IUnknown`, …) do not list `IComIID` in their base list, so generic helpers built around `where T : unmanaged, IComIID` cannot accept them.

This forces consumers to hand-author two polyfills per project:

1. An instance-based `IComIID` interface declaration gated `#if !NET`.
2. A `partial struct Foo : IComIID { Guid IComIID.Guid => IID_Guid; }` entry for **every** generated COM struct that flows through `ComScope<T>`, gated `#if !NET`.

Each new COM type added to `NativeMethods.txt` requires a corresponding partial. The pattern is mechanical and lends itself to generator support.

## Steps to reproduce

Create the project as described below and build it.

The `net10.0` build succeeds. The `net472` build fails with:

```
error CS0246: The type or namespace name 'IComIID' could not be found
error CS0311: The type 'IRunningObjectTable' cannot be used as type parameter 'T' in
              the generic type or method 'Program.GetIID<T>()'. There is no implicit
              reference conversion from 'IRunningObjectTable' to 'IComIID'.
```

## Expected

CsWin32 emits `IComIID` and attaches it to generated COM structs on all supported target frameworks, in a shape appropriate for the language version (static-abstract on .NET 7+, instance member on older targets, matching how `dotnet/winforms` polyfills the type today).

## Actual

The interface and its `partial` attachments are only emitted on TFMs where `static abstract` interface members are supported, leaving older targets without any substitute.

## Environment

| Component | Version |
|---|---|
| `Microsoft.Windows.CsWin32` | `0.3.275` |
| .NET SDK | 10.0.100 |
| TFMs in repro | `net10.0;net472` |

## How the workaround looks in production

[`dotnet/msbuild`](https://github.com/dotnet/msbuild) multi-targets `net10.0` and `net472` and currently maintains two hand-written polyfill files:

[...]

PR #1706, opened 2026-05-27
Mark internal COM struct wrappers `[CLSCompliant(false)]` when they carry CCW thunks

### LINKED ISSUE #1703: CsWin32-generated COM structs trip `CS3016` under `[assembly: CLSCompliant(true)]`
## Summary

When the consuming assembly is marked `[assembly: CLSCompliant(true)]`, every CsWin32-generated COM struct that decorates a method parameter with `[MarshalAs(UnmanagedType.SafeArray, SafeArraySubTypes = new[] { ... })]` produces

```
warning CS3016: Arrays as attribute arguments is not CLS-compliant
```

Examples in the generated surface include `ITypeInfo`, `ITypeLib`, `IRecordInfo`, `IStream`, `IClassFactory`, etc. Under `TreatWarningsAsErrors=true` (the default for official builds in many repos, including dotnet/msbuild) these become hard errors.

The warnings cannot be cleanly suppressed:

- `[SuppressMessage]` in a global suppressions file does **not** work for `CSxxxx` warnings — the C# compiler only honors `SuppressMessageAttribute` for analyzer diagnostics. See <https://github.com/dotnet/roslyn/issues/68526>.
- A blanket `<NoWarn>CS3016</NoWarn>` masks legitimate user code violations.
- The generated structs are `internal`, so `[CLSCompliant(false)]` is semantically meaningless on them — but applying it via a hand-written `partial` declaration is currently the only targeted way to silence the diagnostic, at the cost of then
  having to suppress `CS3019` ("CLS compliance checking will not be performed … because it is not visible from outside this assembly") for that same folder.

## Expected

CsWin32-generated COM struct wrappers should not produce CLS-compliance warnings in a CLS-compliant assembly. They are `internal` and not part of the assembly's CLS surface; the generator could mark them `[CLSCompliant(false)]` (or omit the
`SafeArraySubTypes` array, or emit the structs in a form the CLS rules accept) so consumers do not have to maintain a hand-written `[CLSCompliant(false)]` partial for each generated COM type.

## Actual

Each generated COM struct that has a `SafeArraySubTypes`-decorated parameter emits `CS3016` at every such call-site declaration. Consumers must either:

- disable `CS3016` globally (loses real diagnostic value for user code), or
- maintain a per-struct `[CLSCompliant(false)] internal partial struct Foo { }` declaration alongside an `.editorconfig` rule that suppresses the resulting `CS3019` noise.

## Environment

| Component | Version |
|---|---|
| `Microsoft.Windows.CsWin32` | `0.3.275` |
| .NET SDK | 10.0.100 |
| TFMs in repro | `net10.0;net472` |

## How the workaround looks in production

[`dotnet/msbuild`](https://github.com/dotnet/msbuild) carries a dedicated file, `src/Framework/Windows/Win32/GeneratedInteropClsCompliance.cs`, that exists solely to attach `[CLSCompliant(false)]` to each generated COM struct that trips `CS3016`, plus an `.editorconfig` rule that suppresses the resulting
`CS3019` warnings for files under the CsWin32-generated namespaces.

### References

- msbuild's hand-written partial file:
  <https://github.com/dotnet/msbuild/blob/main/src/Framework/Windows/Win32/GeneratedInteropClsCompliance.cs>
- Roslyn limitation that blocks `[SuppressMessage]` for `CSxxxx`:
  <https://github.com/dotnet/roslyn/issues/68526>

## Suggested fix

Emit `[CLSCompliant(false)]` on generated COM struct wrappers that contain non-CLS-compliant attribute arguments (or otherwise structure the generated `[MarshalAs(...)]` usage so the array argument is not required). Either change removes the need for consumers to author per-struct partial declarations.

## Sources

Drop these four files into a fresh folder and run `dotnet build`.

### `ClsCompliantSafeArray.csproj`

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFrameworks>net10.0;net472</TargetFrameworks>
    <OutputType>Exe</OutputType>
    <LangVersion>latest</LangVersion>
    <Nullable>enable</Nullable>
    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="0.3.275" PrivateAssets="all" />
  </ItemGroup>
</Project>
```

### `NativeMethods.txt`

```
ITypeInfo
ITypeLib
IRecordInfo
```

### `NativeMethods.json`

```json
{ "$schema": "https://aka.ms/CsWin32.schema.json", "allowMarshaling": false, "useSafeHandles": false }
```

### `Program.cs`

```csharp
using System;

[assembly: CLSCompliant(true)]

internal static class Program
{
    // The mere presence of the generated ITypeInfo / ITypeLib / IRecordInfo
    // structs surfaces CS3016. No call site is required.
    private static void Main() => Console.WriteLine("see warnings above");
}
```

## Sketch of the fix

In the CsWin32 generator, when emitting a COM struct wrapper that contains any parameter decorated with
`[MarshalAs(UnmanagedType.SafeArray, SafeArraySubTypes = new[] { ... })]` (or any other attribute whose argument is an array), also emit `[CLSCompliant(false)]` on the struct declaration itself:

```csharp
[CLSCompliant(false)] // <-- new
internal partial struct ITypeInfo
{
    public unsafe void Invoke(
        ...,
        [MarshalAs(UnmanagedType.SafeArray, SafeArraySubTypes = new[] { typeof(...) })]
        object[] rgvarg,
        ...);
}
```

Because the structs are `internal`, this attribute is semantically a no-op for the assembly's CLS surface but suppresses `CS3016` at the source. The generator should **not** also need to suppress `CS3019` — that warning only fires when the
attribute is applied to a member that the compiler thinks isn't externally visible *and* the assembly is CLS-compliant. Suppression of `CS3019` for generated files remains the consumer's responsibility (e.g. via an
`.editorconfig` rule scoped to the generated namespace), but is one-line and stable across new COM types.

Alternative: emit `SafeArraySubTypes` as a single-valued attribute argument when only one subtype is needed, or omit it entirely when the generator can prove it is unnecessary for blittable signatures (`allowMarshaling: false`).

JeremyKuhne:
Fixes #1703.

## Summary

Generated COM struct wrappers contain CCW thunks decorated with
`[UnmanagedCallersOnly(CallConvs = new[] { typeof(CallConvStdcall) })]` — an array-valued attribute argument that trips `CS3016 "Arrays as attribute arguments is not CLS-compliant"` whenever the consuming assembly is marked `[assembly: CLSCompliant(true)]`. Under `TreatWarningsAsErrors=true` (the default for official builds in many repos, including dotnet/msbuild) these become hard build errors. The only targeted workaround today is a hand-authored `[CLSCompliant(false)]` partial per generated COM struct plus an `.editorconfig` rule that suppresses the resulting `CS3019` warnings — see [dotnet/msbuild's `GeneratedInteropClsCompliance.cs`](https://github.com/dotnet/msbuild/blob/main/src/Framework/Windows/Win32/GeneratedInteropClsCompliance.cs).

## Fix

The generator now emits `[CLSCompliant(false)]` on each generated COM struct wrapper, but only when **both** of the following hold:

1. **The struct has CCW thunks** (`ccwThisParameter is not null`). The `UnmanagedCallersOnly` thunks are the only source of an array-valued attribute argument, and they are only emitted on TFMs where `canUseUnmanagedCallersOnlyAttribute` is true (.NET 5+). On `net472` / `netstandard2.0` the generator produces no such attribute, so `[CLSCompliant(false)]` would be unmotivated noise there.
2. **The struct's visibility is `internal`** (the generator's default). For public wrappers the consumer owns the CLS-compliance contract of their public surface and the generator must not unilaterally stamp it.

`CS3019` and `CS3021` (the two downstream "CLS attribute on internal type" diagnostics that fire on these same structs depending on whether the consuming assembly opts into CLS compliance) are added to the generated-file `#pragma warning disable` list so the consumer needs no per-namespace `.editorconfig` overrides — they see zero warnings regardless of their `[assembly: CLSCompliant]` choice.

## Tests

New theories on `test/Microsoft.Windows.CsWin32.Tests/COMTests.cs`:

- **Positive** — `COMStructWrappers_AreCLSCompliantFalse_Issue1703` on `net8.0`/`net9.0`/`net10.0`: generates the issue's exact `NativeMethods.txt` (`ITypeInfo`, `ITypeLib`, `IRecordInfo`), asserts every CCW-bearing COM struct wrapper bears `[CLSCompliant(false)]`, and compiles a `[assembly: CLSCompliant(true)]` consumer **clean of `CS3016`, `CS3019`, and `CS3021`** with no diagnostic filter.
- **Negative (downlevel TFMs)** — `COMStructWrappers_NoCLSCompliantFalse_OnDownlevelTFMs_Issue1703` on `net472`/`netstandard2.0`: no CCW thunks are emitted, so the wrapper must **not** carry `[CLSCompliant(false)]`.
- **Negative (public visibility)** — `COMStructWrappers_NoCLSCompliantFalse_WhenPublic_Issue1703` with `GeneratorOptions.Public = true`: the wrapper is public, so the generator must **not** stamp `[CLSCompliant(false)]` and override the consumer's CLS-surface contract.

## Manual repro verification

Built the repro from #1703 (`TargetFrameworks=net10.0`, `[assembly: CLSCompliant(true)]`, `TreatWarningsAsErrors=true`):
- Against published `0.3.275`: ~70 `error CS3016` from generated CCW thunks ✅ (reproduced)
- Against locally-packed build of this branch: `Build succeeded. 0 Warning(s) 0 Error(s)` ✅ (fixed, no consumer-side workaround)

## End-to-end validation against dotnet/msbuild

A combined local package was consumed in `dotnet/msbuild` to verify the fix in production conditions. With this build:

- `src/Framework/Polyfills/IComIID.cs` and `src/Framework/Polyfills/IComIIDPolyfills.cs` (the #1704 polyfills — addressed by #1705) can be deleted.
- `src/Framework/Windows/Win32/GeneratedInteropClsCompliance.cs` (the hand-authored `[CLSCompliant(false)]` partials this PR replaces) can be deleted.
- The corresponding `.editorconfig` rule that suppresses `CS3019` for generated CsWin32 namespaces can be deleted.

The official-build `TreatWarningsAsErrors=true` build remains clean.


JeremyKuhne:
Ugh, this isn't done. Over-eager agent.

jevansaks:
> For public wrappers the consumer owns the CLS-compliance contract of their public surface and the generator must not unilaterally stamp it.

How does this interact with the extensionReceiver part? If the extensionReceiver is public, do the contributors need to be public too so this check "just works" even in that mode?

jevansaks (approved):

JeremyKuhne:
> How does this interact with the extensionReceiver part? If the extensionReceiver is public, do the contributors need to be public too so this check "just works" even in that mode?

It shouldn't interact with it.

PR #1718, opened 2026-06-10
Add integration tests to PR gate and fix Roslyn 4 analyzer leg

jevansaks:
﻿## Summary

Closes the integration test gap in PR gates and fixes a regression from #1701.

### 1. Fix Roslyn 4 analyzer embedded resources (regression from #1701)

The `Microsoft.Windows.CsWin32.Roslyn4` sibling project includes template `.cs` files as `EmbeddedResource` from the shared source root. Since these files are outside the project cone, MSBuild dropped the `templates/` directory prefix from the resource names, producing `Microsoft.Windows.CsWin32.PInvokeClassHelperMethods.cs` instead of the expected `Microsoft.Windows.CsWin32.templates.PInvokeClassHelperMethods.cs`.

This caused the `Generator` static constructor to throw `GenerationFailedException: Missing embedded resource`, wrapped in a `TypeInitializationException`, which broke **all** consumers using the Roslyn 4.11 analyzer leg (including non-SDK projects and hosts with Roslyn < 5.0).

**Fix:** Add `Link` metadata to the `EmbeddedResource` items to preserve the `templates\` directory structure in resource names.

### 2. Analyzer leg deduplication for non-SDK projects

Non-SDK projects (and hosts with older NuGet tooling) don't understand the `roslyn{version}` folder convention introduced in NuGet 6.12. They pass ALL analyzer DLLs from ALL sub-folders as `/analyzer` references to csc, which causes both the `roslyn4.11` and `roslyn5.0` source generators to run simultaneously, producing duplicate generated code.

**Fix:** Add an MSBuild target (`_CsWin32DeduplicateAnalyzerLegs`) in the package's `.targets` file that removes the `roslyn4.11` leg when `roslyn5.0` analyzers are also present, ensuring only one source generator instance runs.

### 3. Add integration tests to GitHub Actions PR gate

The integration tests (which pack the nupkg and build test projects against it to validate package structure) were only running in the AzDO outer loop pipeline. This PR adds an `integration-test` job to `.github/workflows/build.yml` that runs on every PR:
- MSBuild non-SDK style (net472)
- MSBuild SDK style
- `dotnet build` SDK style
- MSBuild buildtask-SDK style
- `dotnet publish` buildtask-SDK style

The `validate` gate job now also requires the integration tests to pass.

### DIFF
--- a/integration-tests/sdk-roslyn5/app/app.csproj
+++ b/integration-tests/sdk-roslyn5/app/app.csproj
@@ -1,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net10.0-windows</TargetFramework>
+    <LangVersion>14</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CsWin32">
+      <Version>$(NuGetPackageVersion)</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="System.Memory" Version="4.6.0" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.1.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\base\base.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/integration-tests/sdk-roslyn5/app/app.csproj
+++ b/integration-tests/sdk-roslyn5/app/app.csproj
@@ -7,11 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Windows.CsWin32">
-      <Version>$(NuGetPackageVersion)</Version>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
-      <PrivateAssets>all</PrivateAssets>
-    </PackageReference>
+    <PackageReference Include="Microsoft.Windows.CsWin32" Version="$(NuGetPackageVersion)" PrivateAssets="all" />
     <PackageReference Include="System.Memory" Version="4.6.0" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="6.1.0" />
   </ItemGroup>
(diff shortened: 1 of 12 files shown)


integration-tests/sdk-roslyn5/app/app.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <TargetFramework>net10.0-windows</TargetFramework>
+    <LangVersion>14</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Windows.CsWin32">
+      <Version>$(NuGetPackageVersion)</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
Sergio0694: Nit: can simplify this:

```suggestion
    <PackageReference Include="Microsoft.Windows.CsWin32" Version="$(NuGetPackageVersion)" PrivateAssets="all" />
```

JeremyKuhne (approved):

Sergio0694 (approved):

Sergio0694 (approved):

E021ntox:
Could you consider designing and shipping a set of built-in analyzer rules with CsWin32 out of the box?

I recently ran into a issue in a complex codebase where a CsWin32 API call produced unexpected behavior. After hours of troubleshooting, the root cause turned out to be an API returning an `HRESULT` that was missing a success check `(Succeeded)`.

The real problem is that legacy codebases often contain a massive amount of similar code with missing checks, making manual auditing nearly impossible. It would be incredibly valuable if CsWin32 came pre-configured with practical rules to automatically warn developers: "APIs returning HRESULT should check their return values."

This is merely one example. I believe there are numerous other APIs in practice that should emit similar warnings to prompt callers to handle other necessary operations.

Shipping these as pre-defined rules would drastically improve the robustness of Win32 calls in .NET.

@jevansaks

jevansaks:
@e021ntox This is a good suggestion, please file a new issue to request it. Please also provide specifics about the APIs you encountered issues with so we can ensure the analyzer covers your scenario.

E021ntox:
@jevansaks `SHGetImageList<IImageList2>` randomly fails with a high probability. Even if it succeeds, `ppvObj.GetIcon()` also throws exceptions intermittently. I don't know why, there were no such issues back in the .NET Framework era.

jevansaks:
> @jevansaks `SHGetImageList<IImageList2>` randomly fails with a high probability. Even if it succeeds, `ppvObj.GetIcon()` also throws exceptions intermittently. I don't know why, there were no such issues back in the .NET Framework era.

There's not much to go on here. Can you try having Copilot investigate this on your end and provide more details / create a repro?

PR #1727, opened 2026-06-12
Flatten anonymous bitfield sub-properties (#408 phase 2)

### LINKED ISSUE #408: Fields in Anonymous nested structs should be exposed with property getters on the nesting struct
This is ugly:

```cs
SYSTEM_INFO si;
var a = si.Anonymous.Anonymous.wProcessorArchitecture;
```

It should just be

```cs
SYSTEM_INFO si;
var a = si.wProcessorArchitecture;
```

We could do this by adding a property getter/setter that reaches into the nested anonymous struct.

JeremyKuhne:
## Summary

Phase 2 follow-up to #1721. Phase 1 surfaces fields nested in anonymous structs/unions as `[UnscopedRef] ref` accessors on the declaring struct, but for a bitfield-bearing nested struct only the raw backing field was reachable that way. This forwards the **computed bitfield sub-properties** (e.g. `PSAPI_WORKING_SET_EX_BLOCK`''s `Valid` / `ShareCount` / `Win32Protection`) as value `get`/`set` properties on the declaring struct, so callers can read and write them directly as `value.ShareCount` instead of `value.Anonymous.Anonymous.ShareCount`.

## Details

- Each forwarded property delegates to the nested property (`readonly get => this.Anonymous...Prop; set => ... = value;`) and inherits its docs via `<inheritdoc>`.
- The projected type is computed from the **bit width** (1 bit -> `bool`, otherwise the smallest integer of the backing field''s signedness that fits), matching the nested property exactly — so no narrowing occurs even when the backing field is wider (e.g. an `nuint` backing field with `byte`/`ushort` sub-properties).
- The raw backing field continues to be surfaced as a `ref` (Phase 1 behavior unchanged; this is purely additive).
- Only bitfields reached through **anonymous** holders are forwarded; bitfields under a **named** holder are not. Gated on C# 11, like the rest of the feature.

## Tests

- **Syntax** (`StructTests`): forwarding shape + types on `PSAPI_WORKING_SET_EX_BLOCK`, a negative case proving the sibling **named** `Invalid` holder''s bitfields are not forwarded, and the C# 10 gate.
- **Functional** (`GenerationSandbox.Tests`, compiled & executed): the forwarded properties alias the same storage as the nested path; all nine sub-fields **pack without bit overlap** (the exact backing-field bit pattern is asserted); and writes to one field do not disturb its neighbors.
- Validated across **net8.0, net472, and net10.0**.

Fixes #408


JeremyKuhne:
Fixes #408

jevansaks:
> Fixes #408

The "Fixes" has to go in the description I'm pretty sure (which I did by editing yours)

jevansaks (approved):

PR #1770, opened 2026-07-30
Select automatic COM and WinRT output marshalling

jevansaks:
## Summary

- Selects automatic `QI(IInspectable)` detection for recognized COM object outputs.
- Defines default-on behavior with a `NativeMethods.json` opt-out.
- Documents source-generated P/Invoke, generated COM managed call paths, built-in COM, and Native AOT behavior.
- Uses COM identity output for managed implementations; the consuming adaptive projection and generic cast select the requested interface.
- Retains the caller-selected policy as the evaluated alternative and updates the comparison with the final decision.
- Keeps exact-`riid` native-server output, unique COM ownership, and non-canonical output pairs out of scope.

## Validation

Docs-only change.

Implementation: #1771

Generated with Copilot CLI, model: GPT-5.6 Sol.


jevansaks:
Prototype implementation: #1771

PR #1771, opened 2026-07-30
Automatically marshal COM outputs as WinRT

jevansaks:
﻿## Summary

- Automatically projects recognized COM `IID`/`void**` outputs through C#/WinRT when the returned identity implements `IInspectable`, with COM fallback on `E_NOINTERFACE`.
- Enables the behavior by default and adds the `comInterop.autoWinRTMarshalling` opt-out.
- Supports flat P/Invoke, generated COM managed call paths, built-in COM, null output, and Native AOT.
- Preserves existing CsWinRT/source-generated COM RCW identity through `ComWrappers.TryGetComInstance` and classic COM RCW identity through `Marshal.GetIUnknownForObject`.
- Uses generated COM marshalling for `[GeneratedComClass]` CCWs and C#/WinRT marshalling for ordinary managed CCWs.
- Documents the selected architecture under `docs/design`.

## Validation

- Full warning-as-error build and package creation.
- Full non-hardware and hardware-dependent test suites.
- Focused .NET 9 and .NET 10 identity, interface, and round-trip tests for CsWinRT RCWs, source-generated COM RCWs, classic COM RCWs, WinRT CCWs, and generated COM CCWs.
- Fresh package-consumption Native AOT publish and execution of the adaptive marshaller.

Design: #1770

Generated with Copilot CLI, model: GPT-5.6 Sol.


jevansaks:
The first hardware run exposed a test-only environment dependency: SIGDN_NORMALDISPLAY returned win on Windows Server (extensions hidden) instead of win.ini. Commit d26630d1 now asserts the stable SIGDN_FILESYSPATH; the focused test passes in Release on .NET 9 and .NET 10.

src/Microsoft.Windows.CsWin32/templates/ComOrWinRTObjectMarshaller.cs
```diff
@@ -0,0 +1,119 @@
...
+
+	/// <summary>Converts either a COM or Windows Runtime managed object to its COM identity.</summary>
+	public static nint ConvertToUnmanaged(object value)
+	{
+		if (value is null)
+		{
+			return 0;
+		}
+
+		return (nint)global::System.Runtime.InteropServices.Marshalling.ComInterfaceMarshaller<object>
```
Sergio0694: Small note worth pointing out: this will make it so that if you try passing a WinRT-enabled object to a Win32 method, it'll get marshalled via built-in ComWrappers. Then if you later try to pass the same object to a WinRT API, it'll go through the WinRT marshaller. Meaning you'll get two different CCWs for it, with different vtables. I'm not sure how it can be avoided, but just figured we might want to at least explicitly document this.

jevansaks: I'm not sure I understand. Does this ComInterfaceMarshaller path cause a new CCW to be created for the WinRT object? Should we use "is" to test the object for being a WinRT object and use a WInRT marshaller instead? We can do that, I just didn't think it was necessary.

Sergio0694: The problem is that if you call `ComInterfaceMarshaller<T>` to marshal an object that doesn't have a tracked CCW already, it will go through the built-in strategy based `ComWrappers` logic. Which means that the CCW you get depends on the (non deterministic) order in which calls to different marshallers are done.

Suppose you have a type that implements `IDisposable` and some generated COM interface. If you pass it through this method first, it'll just fail, because it's not a `GeneratedComClass`. However if you marshal it through WinRT first, it'll work fine (because the interop generator will detect `IDisposable`, which is custom-mapped, and also harvest the generated COM interface), and then passing it through here later will also work as it'll just reuse the existing CCW (from CsWinRT). This is just an example, there's lots of subtle ways you can get different marshalling results (in some cases you might even end up with multiple CCWs with different vtables in different places), depending on the order in which you pass stuff.

If you need concrete examples you can repro, I can put together some tests/repros or something.
Also FYI @manodasanW in case you have thoughts on the above.

jevansaks: Ok soooo what's your suggestion for how to fix it?

test/GenerationSandbox.BuildTask.Tests/ComOutPtrMarshallingTests.cs
```diff
@@ -0,0 +1,201 @@
...
+using Windows.Win32.System.SystemServices;
+using Windows.Win32.UI.Shell;
+
+namespace GenerationSandbox.BuildTask.Tests;
+
+/// <summary>
+/// Runtime coverage for automatic COM and Windows Runtime output projection.
+/// </summary>
+[Trait("WindowsOnly", "true")]
+public partial class ComOutPtrMarshallingTests
```
jevansaks, line 21: TODO: Add a test for this:
```
      private static void TransferForegroundPrivileges(IWinRTObject provider)
        {
            unsafe
            {
                _ = CoAllowSetForegroundWindow(provider, null);
            }
        }
```

Make sure that the marshalled object works correctly.

jevansaks: Added in 0dc5f017. The runtime test passes a `StorageFile` as `IWinRTObject` to `CoAllowSetForegroundWindow`, then invokes `StorageFile.Name` to verify the object remains usable after marshalling. It intentionally does not require `S_OK`: this in-process fixture does not support foreground transfer, for which the API documents `E_NOINTERFACE`. The test passes on .NET 9 and .NET 10.

src/Microsoft.Windows.CsWin32/templates/ComOrWinRTObjectMarshaller.cs
jevansaks: Implemented in 77283ffe. The marshaller now preserves existing RCWs before selecting a CCW strategy: `ComWrappers.TryGetComInstance` preserves ComWrappers-based RCWs (including CsWinRT and source-generated COM), while `Marshal.IsComObject`/`Marshal.GetIUnknownForObject` preserves classic COM RCWs. New managed objects use `ComInterfaceMarshaller<object>` only when the runtime type is directly marked `[GeneratedComClass]`; all other managed objects use `WinRT.MarshalInspectable<object>`.

Focused tests now pin canonical `IUnknown` identity, interface availability, and round-trip behavior for CsWinRT RCWs, source-generated COM RCWs, classic COM RCWs, WinRT CCWs, and generated COM CCWs. This avoids creating a second CCW for an existing RCW and makes the CCW strategy deterministic for new objects.

test/Microsoft.Windows.CsWin32.Tests/COMTests.cs
```diff
@@ -562,6 +562,175 @@ public void ComOutPtrGenericOverload_IMoniker_BindToObject(bool allowMarshaling)
...
+        Assert.Contains("if (hr != E_NOINTERFACE)", helper, StringComparison.Ordinal);
+        Assert.Contains("ComInterfaceMarshaller<object>.ConvertToManaged", helper, StringComparison.Ordinal);
+        Assert.Contains("ComWrappers.TryGetComInstance", helper, StringComparison.Ordinal);
+        Assert.Contains("Marshal.IsComObject", helper, StringComparison.Ordinal);
+        Assert.Contains("Marshal.GetIUnknownForObject", helper, StringComparison.Ordinal);
+        Assert.Contains("GeneratedComClassAttribute", helper, StringComparison.Ordinal);
+        Assert.Contains("global::WinRT.MarshalInspectable<object>.FromManaged(value)", helper, StringComparison.Ordinal);
+        Assert.Contains(".ConvertToUnmanaged(value)", helper, StringComparison.Ordinal);
+        Assert.DoesNotContain("IidMarshaller", helper, StringComparison.Ordinal);
+        Assert.DoesNotContain("requestedIids", helper, StringComparison.Ordinal);
```
Sergio0694: This test is pretty difficult to read, couldn't we just compare the actual exact generated code?

Sergio0694 (approved):
LGTM, seems very convenient!

test/Microsoft.Windows.CsWin32.Tests/COMTests.cs
jevansaks: Yeah gonna clean this up...

Sergio0694 (approved):

PR #1822, opened 2026-09-16
Fix IInspectable-derived COM interface layout

### LINKED ISSUE #1601: WinRT interfaces should inherit IInspectable otherwise vtable and vfn indices are incorrect
When generating WinRT interop types, interfaces that represent WinRT types should inherit **IInspectable**. For example:

```csharp
partial interface IActivationFactory : IInspectable
```

If the generated interface does not inherit IInspectable, the resulting vfn indices are wrong, which breaks COM/WinRT calls and causes runtime failures.

```
    <PublishAot>true</PublishAot>
```

jevansaks:
Fixes #1601

## Summary
- Generate the metadata-defined `IInspectable` interface as an IUnknown-compatible base, aligning inherited vtable slots in source-generated and built-in COM. Resolve inherited method metadata through its owning generator for cross-WinMD cases.
- Support runtime-class outputs such as `CreateFromWICBitmap(..., out SoftwareBitmap bitmap)` by requesting the class's default interface IID from `ProjectedRuntimeClassAttribute`. Parameterized default interfaces retain C#/WinRT's IID calculation.
- Keep adaptive object projection unchanged. Built-in COM callers explicitly adapt the projected factory with `.As<ISoftwareBitmapNativeFactory>()`; the earlier wrapper-selection workaround has been removed.
- Keep the `out object` regression and add `out SoftwareBitmap` coverage in both COM modes. Add focused IID checks for runtime classes, parameterized default interfaces, and rejection of legacy property-name-based projections.

## NativeAOT
The default-interface lookup requires modern, type-based C#/WinRT projection metadata. A narrowly scoped `IL2072` suppression follows C#/WinRT's documented distinction between modern GUID metadata and its legacy public-field reflection fallback; legacy property-name-based projections are rejected rather than reflected over.

Manually published and executed a Windows x64 .NET **10.0.12 NativeAOT** repro using `CsWin32RunAsBuildTask=true`, source-generated COM, `DisableRuntimeMarshalling=true`, and warnings treated as errors. The executable confirmed `RuntimeFeature.IsDynamicCodeSupported == false` and successfully created and used writable 2x3 BGRA8 bitmaps through both `out SoftwareBitmap` and `out object`. Runtime-class, parameterized-default-interface, ordinary COM, and object IID selection were also exercised.

## Testing
With `NBGV_GitEngine=Disabled` and warnings treated as errors:
- Source-generated `ComOutPtrMarshallingTests`: **18 passed on .NET 9 and 18 on .NET 10**.
- Built-in `AutoWinRTMarshallingBuiltInComTests`: **4 passed on .NET 9 and 4 on .NET 10**.
- Core generator adaptive-marshalling, IInspectable-layout, and cross-WinMD checks: **17 passed**.
- Manual NativeAOT publication and execution described above passed.


src/Microsoft.Windows.CsWin32/Generator.Com.cs
```diff
@@ -859,7 +856,7 @@ static ExpressionSyntax ThisPointer(PointerTypeSyntax? typedPointer = null)
         allMethods.AddRange(typeDef.GetMethods().Select(methodHandle => new QualifiedMethodDefinitionHandle(this, methodHandle)));
 
         AttributeSyntax ifaceType = InterfaceType(
-            foundIInspectable ? ComInterfaceType.InterfaceIsIInspectable :
+            foundIInspectable ? ComInterfaceType.InterfaceIsIUnknown :
```
Sergio0694, line 859: I'm not super familiar with CsWin32's generation pipeline, but this like seems odd. Is this because the COM generator doesn't support `IInspectable`? How do the vtable slot align in this case though? I guess perhaps adding a brief comment above this would help make things clearer.

Sergio0694 (approved):

src/Microsoft.Windows.CsWin32/Generator.Com.cs
jevansaks: Correct, the source generated COM doesn't support that. Good suggestion to leave a comment.

